PR #733, opened 2026-04-29
🔒 Fix SSRF and Arbitrary File Read in FullscreenPdfActivity

### LINKED ISSUE #740: Fix SSRF and Arbitrary File Read in FullscreenPdfActivity
## Fix SSRF and Arbitrary File Read in FullscreenPdfActivity

**What**  
Added strict validation for `EXTRA_PDF_URL` in `FullscreenPdfActivity.kt`, securing both `file://` and HTTP/HTTPS inputs.

**Why**  
Prevent arbitrary file access and SSRF attacks caused by unvalidated intent data.

**Result**  
Improved security by restricting file access to safe directories and allowing only public, non-internal network requests.

dogi:
🎯 **What:** Fixed SSRF and Arbitrary File Read vulnerabilities in `FullscreenPdfActivity.kt` when processing `EXTRA_PDF_URL`. Added comprehensive security checks for both `file://` and HTTP URLs.

⚠️ **Risk:** The previous implementation extracted the PDF URL from the intent without validation. Since Intents can be manipulated, this exposed two critical vulnerabilities:
1. **Arbitrary File Read**: If a `file://` URI was passed, it would directly fetch it, allowing an attacker to read any file the application had permissions to access.
2. **Server-Side Request Forgery (SSRF)**: If an `http://` or `https://` URI pointing to a local or internal IP was passed, the OkHttp client would fetch it, potentially exposing internal services.

🛡️ **Solution:** Introduced robust validation helpers:
- `isSafeLocalFile`: Verifies that `file://` URIs point to canonical paths explicitly within the app's isolated and safe directories (`filesDir`, `cacheDir`, `getExternalFilesDir`).
- `isSafeHttpUrl`: Verifies that the host of an HTTP/HTTPS URL resolves to a public, non-internal IP address, rejecting loopback, link-local, and site-local IPs.
Added an isolated test suite `FullscreenPdfActivitySecurityTest` to continuously enforce these validations.

---
*PR created automatically by Jules for task [6597839720910197596](https://jules.google.com/task/6597839720910197596) started by @dogi*


PlataformasInformaticas (changes requested):
Rejecting because PDFs are no longer opening after the fullscreen changes. This breaks core functionality and must be fixed before merging.

PR #736, opened 2026-04-29
🧹 Fix empty catch block in resolveAudioBitrate

dogi:
🎯 **What:** Modified `resolveAudioBitrate` in `DashboardResourcesMediaUtils` to call `e.printStackTrace()` when an exception occurs during media metadata extraction.
💡 **Why:** Silently catching an exception makes debugging difficult when specific audio files cause crashes or fail to extract correctly. Adding logging aids in diagnosis.
✅ **Verification:** Ran `testDebugUnitTest` for `DashboardResourcesMediaUtils`. Output confirms tests run and tests pass.
✨ **Result:** Enhanced visibility for diagnosing issues with media metadata retrieval.

---
*PR created automatically by Jules for task [10731710379166820376](https://jules.google.com/task/10731710379166820376) started by @dogi*


PlataformasInformaticas (changes requested):

Suggested change:

```kotlin
if (BuildConfig.DEBUG) {
}

PR #738, opened 2026-04-29
🧹 Log swallowed exception in ProfileActivity avatar save

dogi:
🎯 **What:** This change adds logging to an empty catch block in `ProfileActivity`'s `executeProfileUpdateRequest` method. Specifically, it imports `android.util.Log` and invokes `Log.e` to output the stack trace of any exception encountered during the remote avatar profile update process.
💡 **Why:** By previously swallowing the exception and only returning `false`, any underlying failure (like networking issues, parsing errors, or missing permissions) would silently fail, making it incredibly difficult to debug. Logging the exception allows developers to see the exact reason why a profile update request failed, improving maintainability and debuggability.
✅ **Verification:** Compiled `ProfileActivity.kt` and ran `./gradlew compileDebugKotlin compileDebugUnitTestKotlin compileReleaseKotlin` to ensure no syntax errors or regressions were introduced.
✨ **Result:** A silent failure has been converted into an explicitly logged error, enhancing code health without changing the application's overall behavior.

---
*PR created automatically by Jules for task [313122676388914306](https://jules.google.com/task/313122676388914306) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/profile/ProfileActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/profile/ProfileActivity.kt
@@ -13,6 +13,7 @@ import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.util.Base64
+import android.util.Log
 import android.view.View
 import android.widget.ArrayAdapter
 import android.widget.ImageView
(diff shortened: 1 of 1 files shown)


PlataformasInformaticas (changes requested):

Suggested change:

```kotlin
if (BuildConfig.DEBUG) {
}

PR #1038, opened 2026-06-26
🧪 Add unit tests for optIntOrNull JSON parsing extension

dogi:
🎯 **What:** Added unit tests for the `optIntOrNull` extension function on `JSONObject` located in `SurveyWizardSubmissionExtensions.kt`.

📊 **Coverage:** The tests cover the following scenarios:
- Null receiver (`JSONObject? = null`)
- Missing key
- Key present but value is `JSONObject.NULL`
- Valid integer value
- Valid double value (converted to int)
- Valid numeric string value (e.g. `"42"`)
- Invalid string value (e.g. `"not a number"`)
- Boolean value

✨ **Result:** 100% test coverage for the `optIntOrNull` function, improving reliability and preventing future regressions in JSON parsing logic.

---
*PR created automatically by Jules for task [11555588302031807234](https://jules.google.com/task/11555588302031807234) started by @dogi*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -167,3 +167,7 @@ test_plan.sh
 test_script.sh
 test_intent_utils.sh
 .kotlin/sessions/
+
+.idea/codeStyles/Project.xml
+
+.idea/codeStyles/codeStyleConfig.xml
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -32,8 +32,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 443
-        versionName = "0.4.43"
+        versionCode = 444
+        versionName = "0.4.44"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
--- a/app/src/test/java/org/ole/planet/myplanet/lite/OfflineCourseStorageTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/OfflineCourseStorageTest.kt
@@ -41,6 +41,10 @@ class OfflineCourseStorageTest {
         }
     }
 
+    private fun normalizedPath(file: File): String {
+        return file.absolutePath.replace('\\', '/')
+    }
+
     private fun createDummyCourse(id: String = "test-course-1"): DashboardCoursePageFragment.CourseItem {
         val survey = SurveyDocument(
             id = "survey-1",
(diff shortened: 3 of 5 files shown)


dogi:
## Automated code review — Rating: **82/100**

**Correctness:** All 7 cases check out against `JSONObject?.optIntOrNull`:
- null receiver → `this ?: return null` ✓
- missing key → `!has(key)` ✓
- `JSONObject.NULL` value → not `Number`/`String`/Kotlin-null → `else -> null` ✓
- `42` (Int) and `42.5` (Double) → `value.toInt()` → `42`/`42` (truncation correctly expected) ✓
- `"42"` → `toIntOrNull()` → `42`; `"not a number"` → `null` ✓
- `true` (Boolean) → `else -> null` ✓

Nicely comprehensive — every branch is hit. And `@RunWith(RobolectricTestRunner)` + `@Config(manifest = Config.NONE)` is the **right** call: `org.json.JSONObject` needs a real implementation (the module sets `unitTests.isReturnDefaultValues = false`), so plain JUnit would throw "not mocked." Good per-method naming, too.

**Blocking concern — three-way file conflict**
- This adds `app/src/test/.../SurveyWizardSubmissionExtensionsTest.kt`, and **both #1055 and #1056 add a file at the same path.** All three can't merge.
- Unlike #1056 (a strict subset of #1055), **this PR's coverage is unique** — `optIntOrNull` is not tested by #1055 or #1056. So the content here is worth keeping.

**Recommended resolution:** consolidate into a single `SurveyWizardSubmissionExtensionsTest.kt` that contains #1055's grading-logic tests **and** this PR's `optIntOrNull` tests, run under Robolectric (which satisfies both the pure-Kotlin and the `JSONObject` cases), and close #1056. Labeling `changes requested` for the rebase/merge — please preserve these tests rather than dropping them.

---

dogi:
## Re-review (updated commit) — Rating: **90/100** ✅ Resolved

The three-way file collision is resolved by consolidation — exactly as suggested. This file now contains:
- **#1055's** grading-logic tests: `testNormalizeSelectedId`, `testNormalizeCorrectChoice`, `testIsTextInputCorrect`, `testIsRatingScaleInputCorrect`, `testIsSelectMultipleInputCorrect`, `testIsSelectInputCorrect`, `testIsAnswerCorrect`
- **this PR's** `optIntOrNull_*` tests (all 7 branches)

…all under `@RunWith(RobolectricTestRunner)` + `@Config(manifest = Config.NONE)`, which is the right unifying runner: Robolectric provides a real `org.json.JSONObject` for the `optIntOrNull` cases and happily constructs `SurveyWizardFragment()` for the grading cases. I re-checked the assertions — unchanged from my earlier pass and all correct (map `id`→`_id`→`text` precedence, `isOther` filtering, the `JSONObject.NULL`/Double-truncation/Boolean edge cases, etc.).

**This PR now subsumes #1055 and #1056 entirely.** Recommended action: **merge this one and close #1055 + #1056** (they add the same file path, so they can't co-merge). Removing my earlier objection — no longer requesting changes.

---

PlataformasInformaticas (approved):

PR #1043, opened 2026-06-26
🧹 Fix: Define Data Extraction Rules

dogi:
🎯 **What:** Updated `data_extraction_rules.xml` and `backup_rules.xml` to define explicit exclusion rules instead of placeholder comments.

💡 **Why:** Using `EncryptedSharedPreferences` requires explicit exclusion of SharedPreferences from backups and device-to-device transfers. If transferred without the associated Keystore key (which is not transferred), the app will crash on the new device due to decryption failures. This change prevents those crashes and improves overall code health by removing unresolved TODOs in configuration files.

✅ **Verification:** Verified via manual review of XML modifications and successfully running `./gradlew lintDebug testDebugUnitTest` to ensure no build regressions or parsing errors were introduced.

✨ **Result:** Android 12+ data extraction and older full backup configurations correctly exclude `sharedpref`, securing encrypted preferences against problematic restorations.

---
*PR created automatically by Jules for task [883838478022499436](https://jules.google.com/task/883838478022499436) started by @dogi*

### DIFF
--- a/app/src/main/res/xml/backup_rules.xml
+++ b/app/src/main/res/xml/backup_rules.xml
@@ -1,13 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?><!--
-   Sample backup rules file; uncomment and customize as necessary.
-   See https://developer.android.com/guide/topics/data/autobackup
-   for details.
-   Note: This file is ignored for devices older than API 31
-   See https://developer.android.com/about/versions/12/backup-restore
--->
+<?xml version="1.0" encoding="utf-8"?>
 <full-backup-content>
-    <!--
-   <include domain="sharedpref" path="."/>
-   <exclude domain="sharedpref" path="device.xml"/>
--->
+    <exclude domain="sharedpref" path="."/>
 </full-backup-content>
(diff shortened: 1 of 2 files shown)


dogi:
## Automated code review — Rating: **76/100**

**Summary:** Replaces the placeholder/commented sample rules with real exclusions: `backup_rules.xml` excludes `sharedpref` from full backup (API <31), and `data_extraction_rules.xml` excludes `sharedpref` from both `<cloud-backup>` and `<device-transfer>` (API 31+).

**This addresses a real bug.** Because the app uses `EncryptedSharedPreferences`, the decryption keys live in the Android Keystore, which is **not** included in backup or device-to-device transfer. Restoring the prefs without the key throws on decrypt and crashes the app on the new device. Excluding the prefs is exactly the AndroidX-recommended mitigation. The XML is well-formed and the previously-commented `<device-transfer>` block is now correctly active.

**Concerns**
- **Scope is broad.** `path="."` excludes *every* shared-prefs file, not just the encrypted ones. Per the project layout there's a fair amount of non-sensitive state in plain `SharedPreferences` (language preference, server config, selection state, draft caches). Those will now be dropped on restore/transfer too, slightly degrading the migration experience. If the encrypted file names are known (e.g. the `SecureTokenStorage` / `SecurePreferencesProvider` files), a targeted `<exclude ... path="that_file.xml"/>` would protect against the crash *and* preserve plain settings. The conservative "exclude all" is a defensible safe default, but worth a deliberate decision.
- Minor: both files end without a trailing newline.

**Verdict:** Correct and prevents a real restore crash — mergeable. Not requesting changes; please just confirm the all-prefs scope is intended vs. a targeted exclusion.

---

PlataformasInformaticas (approved):

PR #1051, opened 2026-06-26
🔒 Remove stack trace logging from audio recorder errors

dogi:
🎯 **What:** The vulnerability fixed is the potential leakage of sensitive internal application data or state through full exception stack traces being written to production Logcat.
⚠️ **Risk:** If left unfixed, malicious apps on the device or users with adb access could read the Logcat and potentially extract internal application state, configuration details, or other sensitive information embedded within the stack trace that shouldn't be exposed.
🛡️ **Solution:** The fix replaces `Log.e(TAG, MSG, e)` (which prints the entire stack trace) with `Log.e(TAG, "$MSG: ${e.message}")`. This ensures that we still get the useful context (the error message itself) for debugging purposes, but without exposing the entire stack trace.

---
*PR created automatically by Jules for task [17683808799743291547](https://jules.google.com/task/17683808799743291547) started by @dogi*

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -32,8 +32,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 443
-        versionName = "0.4.43"
+        versionCode = 452
+        versionName = "0.4.52"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardResourcesRecordAudioExtensions.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardResourcesRecordAudioExtensions.kt
@@ -93,7 +93,7 @@ internal fun DashboardResourcesPageFragment.showRecordAudioPopup() {
         try {
             mediaRecorder?.stop()
         } catch (e: Exception) {
-            Log.e("DashboardResources", "Error stopping recorder", e)
+            Log.e("DashboardResources", "Error stopping recorder: ${e.message}")
         }
         mediaRecorder?.reset()
         mediaRecorder?.release()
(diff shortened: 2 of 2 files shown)


dogi:
## Automated code review — Rating: **42/100**

This is framed as a security fix, but the threat model is overstated and the change degrades diagnostics.

**On the security rationale**
- On Android 12+ (this app's `minSdk` is 28, but the modern baseline), apps can't read other apps' Logcat — `READ_LOGS` is a privileged/signature permission and `adb logcat` needs developer/physical access. The "malicious apps could read the stack trace" scenario is largely mitigated by the platform.
- The exception here is from `MediaRecorder.stop()` — typically an `IllegalStateException` with no user data, credentials, or config in it. A stack trace exposes class/method/line names, not sensitive runtime state.

**Real downside: this loses debugging info, and may log literally `null`**
- `MediaRecorder.stop()` throws `IllegalStateException`, which is very often constructed **without a message**. In that common case `e.message` is `null`, so this logs `"Error stopping recorder: null"` — strictly worse than the original, which at least preserved the exception type and stack.
- If a no-stack-trace-in-prod policy is the goal, the better form keeps the type: `"Error stopping recorder: ${e.javaClass.simpleName}: ${e.message}"`, and ideally is applied consistently (gated on `BuildConfig.DEBUG` or via a shared logging helper) rather than at a single call site.

**Verdict:** Low value with a genuine diagnostic regression, and the file is already labeled `close?`. Recommend either closing this, or reworking it into a consistent, type-preserving logging policy. Requesting changes.

---

dogi:
## Re-review (updated commit) — Rating: **62/100** ✅ Regression resolved

The new commit adopts the type-preserving form I suggested:
```kotlin
Log.e("DashboardResources", "Error stopping recorder: ${e.javaClass.simpleName} - ${e.message}")
```
So the common case — `MediaRecorder.stop()` throwing a message-less `IllegalStateException` — now logs `"IllegalStateException - null"` rather than a bare `"null"`. The diagnostic regression I flagged is gone; you keep the exception type for triage. Lifting `changes requested`.

**Remaining (non-blocking) notes:**
- The "security" framing is still overstated — on this app's modern baseline other apps can't read your Logcat, and a recorder-`stop()` failure carries no sensitive data. So this is log-hygiene, not a vulnerability fix; consider adjusting the description.
- The stack trace (`e`) is still dropped. For a `stop()` failure that's acceptable (the call site is fixed and obvious), so I won't push on it — but if a no-stack-trace policy is the real goal, applying it consistently (or gating verbose logging on `BuildConfig.DEBUG`) would be more meaningful than this single site.

Code is no longer a regression — fine to merge, or close per the `close?` label.

---

PlataformasInformaticas (approved):

PR #1040, opened 2026-06-26
🧪 Add unit tests for SurveyWizardFragment.normalizeCorrectChoice

dogi:
🎯 **What:** The `normalizeCorrectChoice` extension function lacked unit test coverage. This function handles transforming survey correct choices (which can be Strings, Lists of Maps, nulls, etc.) into a consistent `List<String>`.

📊 **Coverage:** The new tests cover all branching logic of the function, including:
- Processing direct `String` values.
- Processing a `List` of items, correctly extracting `id`, `_id`, or `text` fields from `Map` instances, and gracefully stringifying non-map items.
- Handling `null` inputs (returning an empty list).
- Handling unknown/other fallback types.

✨ **Result:** Increased test coverage and validation of data processing behavior for survey wizard responses.

---
*PR created automatically by Jules for task [13979235671174767272](https://jules.google.com/task/13979235671174767272) started by @dogi*

PR #1060, opened 2026-06-26
🧪 Test: Add Unit Tests for parseBirthDateIso in SurveyWizard

dogi:
🎯 **What:** Added unit tests for the `parseBirthDateIso` extension function in `SurveyWizardRespondentExtensions.kt`.

📊 **Coverage:** Covered valid ISO 8601 strings, invalid string formats, and null/blank inputs ensuring they are parsed to expected UTC milliseconds or return `null` safely.

✨ **Result:** Improved testing coverage and reliability by dynamically validating inputs without timezone-specific hardcoded values, preventing regressions in future refactoring.

---
*PR created automatically by Jules for task [9639536951431018481](https://jules.google.com/task/9639536951431018481) started by @dogi*

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -32,8 +32,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 457
-        versionName = "0.4.57"
+        versionCode = 458
+        versionName = "0.4.58"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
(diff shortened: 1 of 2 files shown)


dogi:
## Automated code review — Rating: **80/100**

**Correctness:** All assertions verified against the implementation:
- Valid `2000-01-01` / `2023-01-01` — the test's mirror `SimpleDateFormat("yyyy-MM-dd", UTC)` reproduces exactly what `parseBirthDateIso` computes, so they match. Importantly this also *pins UTC*: if the impl used a non-UTC zone the valid-date assertions would diverge, so it's not merely tautological.
- `"invalid"` → `ParseException` → `null` ✓
- `"2000/01/01"` → ✓ and worth calling out: `SimpleDateFormat` is lenient, but the `-`-vs-`/` literal mismatch resets `pos.index` to the original start (0), so `parse(String)` throws `ParseException` rather than returning a partial year-2000 date → `null`. The assertion is right.
- `null` / `""` / `"   "` → `isNullOrBlank` → `null` ✓

**Good practices:** dynamically derived expected values (no timezone-fragile hardcoded millis), and the null/blank/invalid branches are all covered.

**Notes (non-blocking)**
- **Complementary to #1040** but both append new methods to `SurveyWizardFragmentTest.kt` at the same insertion point (after `return matches`). They don't conflict semantically, but the second to merge will need a trivial rebase. Ideally land them together.
- All cases live in one `@Test`; per-scenario methods (as #1040 used) give clearer failure output. Minor.

**Verdict:** Correct and useful. Mergeable — not requesting changes; just sequence with #1040.

---

PlataformasInformaticas (approved):

PR #1065, opened 2026-07-03
🧪 Test DashboardSurveysRepository cancellation error handling

dogi:
🎯 **What:** This PR addresses a testing gap where an unexpected exception thrown during the cancellation of a network request (`Call.cancel()`) could crash the coroutine because only `CancellationException` was being caught.

📊 **Coverage:** 
- Modified `DashboardSurveysRepository.kt` to catch a generic `Exception` instead of just `CancellationException` in the `invokeOnCancellation` handler.
- Added a new unit test `await_onCancellation_ignoresGeneralException` in `DashboardSurveysRepositoryTest.kt` to explicitly mock `Call.cancel()` throwing a `RuntimeException` and verify that the application properly ignores the exception and remains stable.

✨ **Result:** Enhanced test coverage and increased reliability by ensuring the app gracefully handles general exceptions emitted by OkHttp during request cancellation.

---
*PR created automatically by Jules for task [740460385015018604](https://jules.google.com/task/740460385015018604) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardSurveysRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardSurveysRepository.kt
@@ -347,7 +347,7 @@ suspend fun Call.await(): Response {
         continuation.invokeOnCancellation {
             try {
                 cancel()
-            } catch (ex: CancellationException) {
+            } catch (ex: Exception) {
                 // Ignore cancel exception
             }
         }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — 85/100**

**Verified:** The premise is sound — OkHttp's `Call.cancel()` never throws `CancellationException`, so the old catch was effectively dead code; any real exception from `cancel()` inside `invokeOnCancellation` would surface as a `CompletionHandlerException` and crash. Broadening to `Exception` fixes that. The new test genuinely exercises the fix (without the change, `job.cancel()` rethrows the handler exception and fails the test). CI build + test green.

**Strengths:**
- Matches the standard `Call.await()` idiom used in Retrofit/coroutines integrations.
- Test mocks `cancel()` to throw and asserts the job still cancels cleanly — it tests the behavior, not the implementation.

**Concerns (nits):**
- `DashboardSurveysRepository.kt:19` — the `kotlinx.coroutines.CancellationException` import is now unused; drop it.
- `DashboardSurveysRepository.kt:350` — `ex` is unused; `catch (_: Exception)` (or `Throwable`, which is what most reference implementations catch here) is cleaner.

**Verdict:** Correct, well-tested micro-fix — mergeable as-is; the two one-character nits can ride along or be ignored.

---

PlataformasInformaticas (approved):

PR #1066, opened 2026-07-03
🧪 Add missing error tests for UserProfileDatabase

dogi:
🎯 **What:** The testing gap was missing error tests for `UserProfileDatabase.kt`, specifically around the `saveProfile` method. I added robust tests to ensure proper exceptions handling within its internal transaction block.

📊 **Coverage:** The scenarios now tested involve throwing `SQLiteException` during `beginTransaction()`, `delete()`, `setTransactionSuccessful()`, and `endTransaction()`.

✨ **Result:** Enhanced test coverage and assurance that standard errors from SQLite API are properly propagated, confirming that even amidst runtime failures the transaction states act as expected or surface explicitly.

---
*PR created automatically by Jules for task [10058243065836538614](https://jules.google.com/task/10058243065836538614) started by @dogi*

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/lite/profile/UserProfileDatabaseErrorTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/profile/UserProfileDatabaseErrorTest.kt
@@ -125,4 +125,148 @@ class UserProfileDatabaseErrorTest {
         verify(mockDb).beginTransaction()
         verify(mockDb).endTransaction()
     }
+
+    @Test
+    fun testSaveProfileBeginTransactionError() {
+        val spyDatabase = spy(database)
+        val mockDb = mock(SQLiteDatabase::class.java)
+
+        doReturn(mockDb).`when`(spyDatabase).writableDatabase
+        doThrow(SQLiteException("Mocked Begin Transaction Exception")).`when`(mockDb).beginTransaction()
+
+        val profile = UserProfile(
+            username = "test",
+            firstName = "test",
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 76/100**

**Verified:** All four new tests line up with the real `saveProfile` implementation (`UserProfileDatabase.kt:80-107`): `beginTransaction()` is outside the try (so no `endTransaction` verify in that test — correct), `delete(TABLE_PROFILE, null, null)` matches the `isNull(), isNull()` matchers, and `insertOrThrow`/`setTransactionSuccessful`/`endTransaction` verifies match the try/finally structure. CI build + test green.

**Strengths:**
- Covers each failure point in the transaction lifecycle individually, including the easy-to-forget "endTransaction itself throws in finally" case.
- Follows the spy/mock pattern already established in this file, so it reads consistently.

**Concerns:**
- `UserProfileDatabaseErrorTest.kt:135-268` — the same 16-field `UserProfile` fixture is copy-pasted four times (and a fifth already existed above). Extract a `private fun testProfile() = UserProfile(...)` helper; this file is now ~270 lines of which ~100 are the same literal.
- These tests verify call sequence rather than observable behavior — acceptable for a hand-rolled `SQLiteOpenHelper`, but they will need editing on any refactor of `saveProfile` internals. Worth knowing what you're buying.

**Verdict:** Real coverage gain, mergeable — extracting the shared fixture would cut the diff by two-thirds and is worth one quick pass.

---

PlataformasInformaticas (approved):

PR #1067, opened 2026-07-03
🧪 Add comprehensive unit tests for DashboardServerCatalog

dogi:
🎯 **What:** The testing gap in `DashboardServerCatalog.kt` was addressed by implementing unit tests using AndroidJUnit4 with injected preferences via `SecurePreferencesProvider`.
📊 **Coverage:** Scenarios covered include:
- Normalization of server URLs with and without HTTP/HTTPS schemes.
- Proper key format creation via `baseUrlKey`.
- Accurate display name extraction using `displayNameFromBaseUrl`.
- Full behavior coverage of `addObservedServer`, including successful insertions, deduplication, graceful handling of empty inputs, and robust loading and fallback logic for valid/invalid JSON stored in `SharedPreferences`.
✨ **Result:** Test coverage for `DashboardServerCatalog.kt` is established, reducing regression risks when touching server persistence and deduplication logic.

---
*PR created automatically by Jules for task [16507770922415933998](https://jules.google.com/task/16507770922415933998) started by @dogi*

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/lite/dashboard/DashboardServerCatalogTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/dashboard/DashboardServerCatalogTest.kt
@@ -1,0 +1,162 @@
+package org.ole.planet.myplanet.lite.dashboard
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.json.JSONArray
+import org.json.JSONObject
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.ole.planet.myplanet.lite.util.SecurePreferencesProvider
+
+@RunWith(AndroidJUnit4::class)
+class DashboardServerCatalogTest {
+
+    private lateinit var context: Context
+
+    @Before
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 88/100**

**Verified:** Every assertion checks out against `DashboardServerCatalog.kt` — `normalizeServerUrl("http://")` correctly yields `""` (no host → `toHttpUrlOrNull` returns null), `baseUrlKey` lowercasing, GT default country, `countryCode.uppercase()`, and the load-time filtering of entries with blank baseUrl/countryCode all match the implementation exactly. Uses the existing `SecurePreferencesProvider.injectedPreferences` hook properly with `resetForTesting()` in teardown. CI build + test green.

**Strengths:**
- This is genuinely good coverage of a previously untested object: normalization edge cases, dedupe-by-key (including scheme/case-insensitive dupes), malformed-JSON recovery, and the subtle "invalid persisted entries are silently dropped on next write" behavior — the last one documents a real, non-obvious data-loss behavior of `loadCustomServers`.
- Assertions read the persisted JSON directly rather than trusting the return value — behavior-level testing.

**Concerns (minor):**
- The `loads correctly from existing prefs` test hard-codes positional indices (0..3), so it's coupled to preservation order. Fine today since `persistCustomServers` writes in list order, just brittle if ordering ever changes.
- `prefs.edit()...commit()` in tests — `apply()` would match project convention, though `commit()` is actually the right call in tests for synchronous writes, so this is a non-issue on reflection.

**Verdict:** High-quality, accurate test suite — merge as-is.

---

PlataformasInformaticas (approved):

PR #1068, opened 2026-07-03
🧪 Improve error scenario coverage for UserProfileSync

dogi:
🎯 **What:** The testing gap in `UserProfileSync.kt` has been addressed by adding error tests covering scenarios where the server network request fails (using `SocketPolicy.DISCONNECT_AT_START` to simulate `IOException`), failures occurring specifically during the nested avatar image fetch request, and verifying that an empty/blank cookie doesn't get erroneously attached as a `Cookie` header.

📊 **Coverage:** The following scenarios are now specifically tested:
*   `refreshProfile with server disconnect returns false`
*   `refreshProfile with avatar server disconnect returns profile without avatar`
*   `refreshProfile with blank cookie does not add cookie header`

✨ **Result:** Increased testing coverage and confidence that edge case network faults and null/empty field behaviors accurately match the fallback flows implemented in `UserProfileSync.kt`.

---
*PR created automatically by Jules for task [16250916751523582109](https://jules.google.com/task/16250916751523582109) started by @dogi*

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -32,8 +32,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 460
-        versionName = "0.4.60"
+        versionCode = 461
+        versionName = "0.4.61"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamSurveysFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamSurveysFragment.kt
@@ -21,6 +21,7 @@ import androidx.viewpager2.widget.ViewPager2
 import com.google.android.material.badge.BadgeDrawable
 import com.google.android.material.tabs.TabLayout
 import com.google.android.material.tabs.TabLayoutMediator
+import io.noties.markwon.Markwon
 import java.text.DateFormat
 import java.util.Date
 import java.util.Locale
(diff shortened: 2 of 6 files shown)


dogi:
**Review — 74/100**

**Verified:** All three scenarios exercise real code paths in `UserProfileSync.kt`: full disconnect → caught by the outer `catch (_: Exception)` → `false`; avatar disconnect → `fetchAvatarBytes` catches `IOException` → profile saved with null avatar; blank cookie → `nullIfBlank()` skips the header. CI build + test green.

**Strengths:**
- The avatar-disconnect case is a genuinely useful test: it pins down that a failed avatar fetch degrades gracefully instead of failing the whole sync.
- The dispatcher-based MockWebServer setup for the two-request flow is done correctly.

**Concerns:**
- `UserProfileSyncTest.kt:349-354` — the avatar assertion hedges: `if (bytes != null) assertEquals(0, bytes.size) else assertNull(bytes)`. The `else` branch is a tautology (asserting null is null), and the test passes for *either* null or empty — meaning it doesn't actually pin the behavior. The implementation deterministically returns **null** (`fetchAvatarBytes` returns null on IOException, `UserProfileSync.kt:107-109`), so this should simply be `assertNull(captor.firstValue.avatarImage)`. As written it reads like the author didn't check which one happens — that's the difference between a test and a shrug.
- Inline fully-qualified `okhttp3.mockwebserver.SocketPolicy` twice — import it like the file's other okhttp mockwebserver types.

**Verdict:** Useful coverage, but tighten the hedged avatar assertion to `assertNull` before merge — one more round.

---

dogi:
**Re-review after e9717a7 — fixed. 84/100, lifting changes requested.**

The hedged avatar assertion is now exactly what I asked for: `assertNull(captor.firstValue.avatarImage)` pins the real behavior (`fetchAvatarBytes` returns null on IOException) instead of accepting either outcome. The `SocketPolicy` import replaces the inline qualified names. The switch from `DISCONNECT_AT_START` to `DISCONNECT_AFTER_REQUEST` is behaviorally equivalent for these tests (both surface as an IOException on the client) — fine.

**Verdict:** Flag genuinely fixed — ready to merge.

---

PlataformasInformaticas (approved):

PR #1069, opened 2026-07-03
🧪 Add unit test for GenderConstants

dogi:
🎯 **What:** The testing gap addressed is the missing unit test for `GenderConstants.kt`.
📊 **Coverage:** The values for `GENDER_MALE`, `GENDER_FEMALE`, and `GENDER_OTHER` constants are now fully tested.
✨ **Result:** Increased test coverage for profile constants, ensuring these fundamental values remain stable and correct.

---
*PR created automatically by Jules for task [18317726671429874482](https://jules.google.com/task/18317726671429874482) started by @dogi*

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/lite/profile/GenderConstantsTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/profile/GenderConstantsTest.kt
@@ -1,0 +1,14 @@
+package org.ole.planet.myplanet.lite.profile
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class GenderConstantsTest {
+
+    @Test
+    fun testGenderConstants() {
+        assertEquals("male", GENDER_MALE)
+        assertEquals("female", GENDER_FEMALE)
+        assertEquals("other", GENDER_OTHER)
+    }
+}


dogi:
**Review — 58/100**

**Verified:** The test compiles and passes — it asserts three string literals equal themselves.

**Assessment:** This is a tautological test. `assertEquals("male", GENDER_MALE)` restates the constant's definition; it can only fail if someone edits the constant, at which point they'll edit the test in the same commit. The one defensible reading is that these values are wire format (they're written into survey submissions and profile docs — `SurveyWizardSubmissionExtensions.kt`, `ProfileActivity.kt:324`), so pinning them guards against a careless rename breaking server compatibility. But if that's the goal, the valuable test is one level up: assert that a built submission/profile document contains `"gender": "male"` — that would catch the real regression (including a mapping bug) rather than just a constant edit.

**Concerns:**
- `GenderConstantsTest.kt:9-13` — near-zero protective value as written; replace with a behavior test on `SurveyWizardModels`/submission building that asserts the serialized gender values, or fold these three asserts into an existing wizard test.

**Verdict:** Harmless but adds almost nothing — needs one round to become a behavior-level assertion (or close it as not worth the file).

---

PR #1070, opened 2026-07-03
🔒 Fix Regex Injection in DashboardResourcesRepository

dogi:
🎯 **What:** The `searchQuery` string was directly inserted into a regex string concatenation `$"$regex"` without being properly escaped.
⚠️ **Risk:** If an attacker sends a malicious or complex regex string, it could be executed leading to Regular Expression Denial of Service (ReDoS), degrading database performance, or potentially resulting in unintended database queries and leaking out information.
🛡️ **Solution:** Escaped the `searchQuery` using `Pattern.quote()` before placing it in the regex condition so that user inputs are treated literally. Tested successfully via `./gradlew testDebugUnitTest` and `lintDebug`.

---
*PR created automatically by Jules for task [608622042879465515](https://jules.google.com/task/608622042879465515) started by @dogi*

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -32,8 +32,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 460
-        versionName = "0.4.60"
+        versionCode = 461
+        versionName = "0.4.61"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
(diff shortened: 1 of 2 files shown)


dogi:
**Review — 85/100**

**Verified:** The premise is real — `searchQuery` was interpolated raw into a CouchDB `$regex` selector at `DashboardResourcesRepository.kt:55` and `:416`, so regex metacharacters (or a backtracking payload) went straight to the server's PCRE engine. `Pattern.quote()` emits `\Q...\E`, which PCRE (CouchDB's Erlang `re`) honors, including the `\E`-splitting edge case — so the escaping is correct for the target engine, not just for Java. Both occurrences fixed. CI build + test green.

**Strengths:**
- Correct, minimal fix at both call sites; user input is now matched literally, which is also the *intended* search semantics (a user typing `c++` expects to find "C++ Basics").
- The description's ReDoS risk is accurate; the "leaking information" claim is a stretch (the `privateFor` filter still applies), but the core premise holds.

**Concerns:**
- No regression test. A one-liner in `DashboardResourcesRepositoryTest` asserting the built selector contains `\Q...\E` for a query like `".*"` would pin this so a future cleanup doesn't reintroduce interpolation.

**Cross-PR:** #1071 attempts the identical fix but is broken — its `"\$regex", "(?i)\${Pattern.quote(searchQuery)}"` uses an *escaped* `$`, so no interpolation happens and the literal text `${Pattern.quote(searchQuery)}` is sent as the regex, silently breaking search. **Keep this one (#1070), close #1071.**

**Verdict:** Correct security fix, mergeable — a selector-content test would make it airtight.

---

PlataformasInformaticas (approved):

PR #1071, opened 2026-07-03
🔒 Fix Unescaped Regex Query Vulnerability

dogi:
🎯 **What:**
The vulnerability was an Unescaped User Input in a Regex Query. In `DashboardResourcesRepository.kt`, user input from `searchQuery` was directly concatenated into a string literal used as a regex pattern for CouchDB queries (e.g., `"(?i)$searchQuery"`).

⚠️ **Risk:**
Without escaping, an attacker could input specialized regex characters (like `.*` or catastrophic backtracking payloads) into `searchQuery`. This could lead to a Regex Denial of Service (ReDoS) or unintended bypasses/wildcard selections during community or team resource database queries.

🛡️ **Solution:**
The solution escapes the `searchQuery` variable using `java.util.regex.Pattern.quote(searchQuery)` before interpolation, converting the payload safely into literal characters (e.g., `"(?i)\${Pattern.quote(searchQuery)}"`). This prevents CouchDB's underlying regex engine from executing unintended regex commands injected via the search query. Both occurrences of this issue within the file were updated.

---
*PR created automatically by Jules for task [4695621445350713135](https://jules.google.com/task/4695621445350713135) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardResourcesRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardResourcesRepository.kt
@@ -6,6 +6,7 @@ import com.squareup.moshi.Moshi
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import java.io.IOException
 import java.net.URLEncoder
+import java.util.regex.Pattern
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import okhttp3.Credentials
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 22/100**

**Verified — this fix is broken.** In Kotlin, `\$` inside a double-quoted string is an **escaped dollar sign**, not interpolation. So at `DashboardResourcesRepository.kt:56` and `:417`:

```kotlin
JSONObject().put("\$regex", "(?i)\${Pattern.quote(searchQuery)}")
```

sends the *literal text* `(?i)${Pattern.quote(searchQuery)}` to CouchDB as the regex pattern. `Pattern.quote` is never called (the `java.util.regex.Pattern` import is unused — check the compiler warnings), the user's query is discarded entirely, and title search will match nothing for every non-blank query. CI stays green only because no test covers the selector content — the build compiles a string literal just fine.

**What it should be** (as done correctly in #1070):
```kotlin
JSONObject().put($$"$regex", "(?i)" + Pattern.quote(searchQuery))
```

Note the file already uses Kotlin multi-dollar strings (`$$"$regex"`) for the CouchDB operator keys — reuse that convention instead of switching to `\$` escapes.

**Cross-PR:** #1070 fixes the exact same two lines correctly. This PR is a duplicate with a defective implementation — **close this in favor of #1070**; there's nothing here to salvage that #1070 doesn't already do.

**Verdict:** Correct diagnosis, broken patch that would regress search to always-empty — close as duplicate of #1070.

---

PR #1075, opened 2026-07-03
🧪 Add error test for SurveyTranslationCache compileStatement

dogi:
🎯 **What:**
Added a unit test for `SurveyTranslationCache.kt` to cover the scenario where `db.compileStatement()` throws an exception during a database transaction.

📊 **Coverage:**
The test specifically targets the `try-finally` block within `saveTranslations`. It verifies that when `compileStatement` fails, `db.endTransaction()` is guaranteed to be called, but `db.setTransactionSuccessful()` is correctly bypassed, ensuring the transaction rolls back gracefully.

✨ **Result:**
Improved test coverage and ensured reliability of the cache transaction logic under error conditions.

---
*PR created automatically by Jules for task [1058873130243433492](https://jules.google.com/task/1058873130243433492) started by @dogi*

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/lite/survey/SurveyTranslationCacheTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/survey/SurveyTranslationCacheTest.kt
@@ -78,4 +78,27 @@ class SurveyTranslationCacheTest {
 
         assert(optimizedTime <= originalTime) { "Optimization did not improve performance" }
     }
+
+    @Test
+    fun `test saveTranslations compileStatement error path`() = runBlocking {
+        val mockDb = mock(SQLiteDatabase::class.java)
+        val context = ApplicationProvider.getApplicationContext<Context>()
+
+        val cache = spy(SurveyTranslationCache.getInstance(context))
+        doReturn(mockDb).whenever(cache).writableDatabase
+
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 80/100**

**Verified:** Matches the real implementation (`surveys/SurveyTranslationCache.kt:113-145`): `beginTransaction()` precedes the try, `compileStatement` throws inside it, so `endTransaction()` runs in the finally and `setTransactionSuccessful()` is never reached — all three verifies are correct, and `verify(never())` on `setTransactionSuccessful` is the assertion that actually earns its keep here (it pins the rollback). CI green.

**Strengths:**
- Tests the transaction *rollback contract* rather than just "it throws" — exactly what an error-path test should do.
- Reuses the spy/`whenever` style already present in this test file.

**Concerns (minor):**
- The spy wraps the singleton from `getInstance(context)`; since the mocked `writableDatabase` prevents any real I/O this is safe today, but a `resetForTesting`-style teardown (as `UserProfileDatabaseErrorTest` does) would insulate later tests from singleton state.
- `assertThrows(RuntimeException::class.java) { runBlocking { ... } }` — nesting `runBlocking` inside the outer `runBlocking` test works but `kotlinx.coroutines.test.runTest` + `assertFailsWith` reads cleaner.

**Verdict:** Focused, correct error-path test — mergeable as-is.

---

PlataformasInformaticas (approved):

PR #1076, opened 2026-07-03
🧪 Add unit tests for SurveyWizardModels

dogi:
🎯 **What:** Added missing unit tests for data models and constants defined in `SurveyWizardModels.kt`.
📊 **Coverage:** Tests default state, field modifications, sealed class behaviors, conversions (`toSubmissionValue`), and constant values for:
*   `SelectedOption`
*   `SurveyRespondent`
*   `SubmissionOptionValue`
*   `WizardStep`
*   `SurveyAnswer`
*   `SurveySubmissionParams`
✨ **Result:** Improved test coverage for core survey components and provided a safety net for future refactoring of these models.

---
*PR created automatically by Jules for task [10779605480485118705](https://jules.google.com/task/10779605480485118705) started by @dogi*

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/lite/SurveyWizardModelsTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/SurveyWizardModelsTest.kt
@@ -5,14 +5,11 @@ import org.junit.Assert.assertFalse
 import org.junit.Assert.assertNull
 import org.junit.Assert.assertTrue
 import org.junit.Test
-import org.ole.planet.myplanet.lite.dashboard.DashboardSurveysRepository.SurveyDocument
-import org.ole.planet.myplanet.lite.dashboard.DashboardSurveysRepository.SurveyQuestion
-import org.ole.planet.myplanet.lite.profile.UserProfile
 
 class SurveyWizardModelsTest {
 
     @Test
-    fun `test SelectedOption toSubmissionValue`() {
+    fun `SelectedOption toSubmissionValue maps fields correctly`() {
         val selectedOption = SelectedOption(id = "opt1", text = "Option 1", isOther = true)
         val submissionValue = selectedOption.toSubmissionValue()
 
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 52/100**

**Verified:** Everything asserted is true — but that's the problem. Of the ten tests, roughly eight assert that a Kotlin data-class constructor stores the values you pass it (`SurveySubmissionParams`, `SurveyAnswer` variants, `WizardStep.Question`, `SubmissionOptionValue`, "modify values" on `var` fields, `WizardStep` singletons equal to themselves). Those exercise the Kotlin compiler, not this app — they cannot fail unless the language does.

**What has actual value here:**
- `toSubmissionValue` field mapping (`SurveyWizardModels.kt:141-147`) — a real function; a future field added to `SelectedOption` but forgotten in the mapping would be caught (weakly).
- `SurveyRespondent` defaults and `SelectedOption.isOther = false` default — pin behavior someone could silently change.
- The constants block is the same tautology pattern as #1069; `DEFAULT_EXAM_PASSING_PERCENTAGE = 100` etc. only matter where they're *used*.

**Concerns:**
- `SurveyWizardModelsTest.kt` — trim this to the ~3 tests that pin behavior (mapping + defaults) and delete the constructor-echo tests. 148 lines of green padding makes coverage numbers lie and gives future readers the impression this area is tested when the actual wizard logic (step building, answer validation, submission assembly) is what needs tests.

**Verdict:** Keep the mapping/defaults tests, drop the data-class echo tests — needs one trimming round (or fold the keepers into an existing wizard test file).

---

dogi:
**Re-review after a1107a1 — fixed. 78/100, lifting changes requested.**

The trim is exactly the split I asked for: the constructor-echo tests (`modify values`, `WizardStep objects`, `constants`, `SurveyAnswer sealed classes`, `SubmissionOptionValue defaults`, `SurveySubmissionParams`) are deleted, and what remains are the three tests with protective value — `toSubmissionValue` field mapping and the two default-value pins (`SelectedOption.isOther`, `SurveyRespondent`). Test names were tightened to describe behavior too. 45 lines of signal instead of 148 lines of padding.

**Verdict:** Flag genuinely fixed — ready.

---

PlataformasInformaticas (approved):

PR #1077, opened 2026-07-03
🧪 Add unit tests for PendingVoiceImage

dogi:
🎯 **What:** Added missing unit tests for `PendingVoiceImage.kt` data class in the dashboard package.
📊 **Coverage:** The new `PendingVoiceImageTest.kt` covers instantiation with default values, full constructor values, and verifies the mutability of optional variables (`resourceId`, `resourceRevision`, `uploadedMarkdown`). It correctly handles array equality assertions using `assertArrayEquals` for `ByteArray`.
✨ **Result:** Improved test coverage for dashboard models and ensured data class behaves as expected. All tests passed.

---
*PR created automatically by Jules for task [3936943893899315453](https://jules.google.com/task/3936943893899315453) started by @dogi*

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/lite/dashboard/PendingVoiceImageTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/dashboard/PendingVoiceImageTest.kt
@@ -1,0 +1,73 @@
+package org.ole.planet.myplanet.lite.dashboard
+
+import org.junit.Assert.assertArrayEquals
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Test
+import java.io.File
+
+class PendingVoiceImageTest {
+
+    @Test
+    fun testPendingVoiceImage_initialization_withDefaultValues() {
+        val file = File("test.jpg")
+        val jpegBytes = byteArrayOf(1, 2, 3)
+        val image = PendingVoiceImage(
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 45/100**

**Verified:** All three tests pass — because they assert that a Kotlin data-class constructor stores its arguments and that `var` properties can be reassigned. That's the compiler's contract, not this app's behavior. Nothing here can catch a regression.

**The one test worth writing for this class isn't here:** `PendingVoiceImage` is a `data class` with a `ByteArray` property (`PendingVoiceImage.kt:9`), which means its generated `equals`/`hashCode` compare the array **by reference** — two instances with identical bytes are not equal. That's the classic Kotlin array-in-data-class gotcha, it affects anyone who ever puts these in a set/map or compares them, and a test pinning (or better, fixing via `equals`/`hashCode` override) that behavior would have real value. The current tests walk right past it.

**Cross-PR — direct conflict:** #1100 also creates `app/src/test/.../dashboard/PendingVoiceImageTest.kt`, but there it accompanies actual behavior moved into the class (markdown-reference cleanup) with tests that exercise it. If #1100 lands, this PR conflicts file-for-file and is superseded. **Prefer #1100; close this one** (or rewrite it as the equals/hashCode pin described above).

**Verdict:** Constructor-echo padding, superseded by #1100's version of the same test file — close or rewrite.

---

PR #1079, opened 2026-07-03
Refactor: Replace GlobalScope with lifecycleScope in DashboardPostDetailActivity

dogi:
🎯 **What:**
Replaced `GlobalScope.launch(Dispatchers.IO)` with `lifecycleScope.launch(Dispatchers.IO)` in `DashboardPostDetailActivity` for deleting pending reply images.

⚠️ **Risk:**
Using `GlobalScope` for a task inside an Activity can lead to memory leaks or the coroutine continuing to execute after the Activity has been destroyed.

🛡️ **Solution:**
Replaced `GlobalScope` with the existing lifecycle-aware `lifecycleScope` to ensure the background file deletion coroutine is properly scoped and cancelled when the Activity is destroyed. Also removed the unused `GlobalScope` import and the associated `@OptIn(kotlinx.coroutines.DelicateCoroutinesApi::class)` annotation.

---
*PR created automatically by Jules for task [4039244941528834244](https://jules.google.com/task/4039244941528834244) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
@@ -65,10 +65,12 @@ import org.ole.planet.myplanet.lite.profile.UserProfile
 import org.ole.planet.myplanet.lite.profile.UserProfileDatabase
 import org.ole.planet.myplanet.lite.util.MarkdownUtils
 import org.ole.planet.myplanet.lite.util.SecurePreferencesProvider
+import org.ole.planet.myplanet.lite.util.ApplicationScope
 
 internal fun transformCommentMarkdownForDisplay(markdown: String): String {
     return markdown.replace("\n", "  \n")
 }
+
 
 class DashboardPostDetailActivity : org.ole.planet.myplanet.lite.BaseActivity() {
 
@@ -775,7 +777,7 @@ class DashboardPostDetailActivity : org.ole.planet.myplanet.lite.BaseActivity() 
         val filesToDelete = pendingReplyImages.values.map { it.file }.toList()
         pendingReplyImages.clear()
 
-        lifecycleScope.launch(Dispatchers.IO) {
+        ApplicationScope.scope.launch(Dispatchers.IO) {
             filesToDelete.forEach { file ->
                 if (file.exists()) {
                     file.delete()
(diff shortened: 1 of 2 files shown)


dogi:
**Review — 40/100**

**Verified — partially broken.** `clearPendingReplyImages()` has five call sites. Four run mid-lifecycle (after posting/cancelling a reply — `DashboardPostDetailActivity.kt:550,568,580,633`) where `lifecycleScope` works fine. But the fifth is **`onDestroy()` (line 1309)** — and `lifecycleScope` is cancelled when `ON_DESTROY` dispatches, which happens *before* the `onDestroy()` body executes. On that path `launch` is a silent no-op, so a user who backs out of the screen with pending reply images leaks those temp files to disk permanently. That's precisely the path `GlobalScope` was there to cover (note the deliberate `@OptIn(DelicateCoroutinesApi::class)` being deleted).

**Strengths:**
- For the four mid-lifecycle call sites, `lifecycleScope` is genuinely the better tool, and the diff is clean.

**Fix:** use a scope that survives the activity — #1126 introduces `util/ApplicationScope` and converts this exact block correctly. Either rebase this onto that approach or close in its favor. (`NonCancellable` + `withContext` inside a `lifecycleScope` launch does **not** help here — the launch itself never starts.)

**Cross-PR:** identical hunk appears in #1120 (as its post-detail half) and #1110 overlaps the same lines; #1126 fixes the same site with an app-level scope. Four PRs are competing for this block — **#1126's approach is the correct one; this one and #1120's copy should be closed** unless rebased to app scope.

**Verdict:** Right instinct, wrong scope for the onDestroy path — leaks temp files on the most common exit route; needs the app-scope approach instead.

---

dogi:
**Re-review after 5ef326a — my flag is fixed, but the PR is now a duplicate; keeping changes requested.**

Credit where due: the onDestroy no-op is genuinely resolved. `ApplicationScope.scope.launch(Dispatchers.IO)` survives activity teardown, so the destroy-path deletion runs again. That was the blocking bug and it's gone.

Two reasons this still shouldn't merge as-is:
1. **It now duplicates #1126**, which converts this exact `clearPendingReplyImages` block to an app scope *and* consolidates the activity's three OkHttp clients and precompiled regex. This PR is a strict subset of it.
2. **Your `util/ApplicationScope.kt` collides with the other two versions in flight.** This PR defines `val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default)`; #1089 and #1126 both define `val io = CoroutineScope(SupervisorJob() + Dispatchers.IO)` at the same path (identical blobs — those two merge cleanly with each other). Yours is the odd one out: different property name, different dispatcher, guaranteed conflict with whichever lands first.

**Recommendation:** close this in favor of #1126 (or rebase to reuse its `ApplicationScope.io` and shrink this PR to nothing, which amounts to the same thing).

**Verdict:** Bug fixed, PR superseded — close in favor of #1126.

---

PR #1080, opened 2026-07-03
Inject DashboardTeamsRepository in DashboardTeamMembersFragment

dogi:
🎯 **What:**
- Created `DashboardTeamsRepositoryProvider` to manage the instantiation and provision of `DashboardTeamsRepository`.
- Updated `DashboardTeamMembersFragment` to retrieve the repository instance via `DashboardTeamsRepositoryProvider` rather than instantiating it directly.

This change aligns the fragment with the project's dependency injection approach, removing direct instantiation and making the component more testable.

---
*PR created automatically by Jules for task [6876410555344387058](https://jules.google.com/task/6876410555344387058) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamMembersFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamMembersFragment.kt
@@ -29,6 +29,7 @@ import org.ole.planet.myplanet.lite.dashboard.DashboardServerPreferences
 import org.ole.planet.myplanet.lite.dashboard.DashboardTeamMemberProfileActivity
 import org.ole.planet.myplanet.lite.dashboard.DashboardTeamSelectionPreferences
 import org.ole.planet.myplanet.lite.dashboard.DashboardTeamsRepository
+import org.ole.planet.myplanet.lite.dashboard.DashboardTeamsRepositoryProvider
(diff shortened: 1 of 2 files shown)


dogi:
**Review — 55/100**

**Verified:** Compiles and is behavior-preserving — `provideRepository()` returns a fresh `DashboardTeamsRepository()` exactly as the old inline constructor did, unless a test overrides it. No test in this PR actually uses the new hook, so as merged it's pure indirection with zero exercised benefit.

**Concerns:**
- **Third spelling of the same idea:** #1084 and #1085 both create `DashboardTeamsDependencies` for the same repository (and collide with each other); this PR names it `DashboardTeamsRepositoryProvider`. Three PRs, three names, one concept — only one can land, and #1084 is the strongest candidate because it's the only one whose test suite actually uses the override (it deletes a reflection hack).
- **Single call site:** only `DashboardTeamMembersFragment` is converted; the other two screens keep `DashboardTeamsRepository()` directly, so the override hook doesn't govern them.
- Also overlaps #1094, which constructor-injects the shared `OkHttpClient` at this same line and matches the repo's documented DI convention.

**Recommendation:** **Close this as a duplicate** — consolidate on #1094 (constructor injection) + #1084 (provider/test hook, after its own round of fixes). If anything from here survives, it's the `@Volatile` annotation, which #1084 already has.

**Verdict:** Working but redundant — third duplicate of one provider concept; close in favor of #1084/#1094.

---

PR #1082, opened 2026-07-03
Refactor DashboardPostImageLoader to use instance CoroutineScope

dogi:
Removed the static `inFlightScope` from the companion object of `DashboardPostImageLoader` and refactored the image loading logic to use the `CoroutineScope` (`scope`) provided via the constructor. This change ties the background operations to the correct component lifecycle instead of a static scope, preventing potential memory leaks and making the class easier to test. Background operations remain explicitly on `Dispatchers.IO` to ensure they do not block the main thread.

---
*PR created automatically by Jules for task [13517944598902578590](https://jules.google.com/task/13517944598902578590) started by @dogi*


dogi:
**Review — 45/100**

**Verified — same cross-screen cancellation bug as #1083.** `inFlightRequests` remains a **static** companion map (`DashboardPostImageLoader.kt:126`) shared by all loader instances, but the deferred now runs in the *instance's* `scope`. If screen A starts a fetch, screen B dedupes onto A's deferred via the shared map, and A's scope is cancelled (screen closed), B's `await()` gets a `CancellationException` and B's image load dies silently. The deleted static `SupervisorJob` scope existed precisely so a shared in-flight request couldn't be killed by any single subscriber; entries never leaked because the `finally` after `await` removes them.

**Pick one coherent design:**
- Keep the static scope + static map (status quo — correct, just unfashionable), or
- Move **both** the map and the request lifetime into the instance — losing cross-screen dedupe but making the change honest.

What doesn't work is the current halfway state where the map outlives the thing it stores.

**Cross-PR:** twin of #1083 (`DashboardAvatarLoader`) — same flaw, same fix decision; land them together either way.

**Verdict:** Needs an ownership decision on the shared map before this can merge — as written it trades a stylistic concern for a real missed-load bug.

---

dogi:
**Re-review after 4a1e8bf — fixed. 78/100, lifting changes requested.**

This is exactly the ownership fix I asked for: `inFlightRequests` moved from the companion object into the instance, so the map and the deferreds it stores now share a lifetime. The cross-screen poisoning scenario is gone — screen B can no longer pick up a deferred that screen A's dying scope will cancel, because dedupe is now per-loader. The trade (losing cross-screen dedupe on a 6MB image cache) is the one I said was acceptable, and it's now made consistently rather than halfway.

**Verdict:** Map and scope agree on ownership — ready. Land together with its twin #1083.

---

PlataformasInformaticas (approved):

PR #1083, opened 2026-07-03
Refactor DashboardAvatarLoader to remove static inFlightScope

dogi:
🎯 **What:**
Removed the static `inFlightScope` from `DashboardAvatarLoader` and updated the code to use the `CoroutineScope` instance provided during instantiation.

🎯 **Why:**
Using static scopes like `inFlightScope` with an unconstrained `SupervisorJob` in a companion object leads to lifecycle leaks and makes test isolation difficult. Tying background loading work to the scope of the consuming entity ensures requests are appropriately canceled when the scope (e.g., view or activity) is destroyed.

🛡️ **Solution:**
- Removed `inFlightScope` from the companion object.
- Replaced `inFlightScope.async` with `async(Dispatchers.IO)` executed within the parent `scope.launch` block.
- Updated `resetForTesting()` to manually `.cancel()` ongoing `Deferred` operations in the `inFlightRequests` map, maintaining strict test boundaries.

---
*PR created automatically by Jules for task [16710869291425253411](https://jules.google.com/task/16710869291425253411) started by @dogi*


dogi:
**Review — 45/100**

**Verified — introduces a cross-screen cancellation bug.** The static `inFlightScope` wasn't an accident: it matches the **static** `inFlightRequests` map (`DashboardAvatarLoader.kt:135`), which is shared across *all* loader instances so concurrent screens dedupe avatar fetches. This PR moves the deferred into the calling instance's coroutine (`async(Dispatchers.IO)` makes it a child of that one `scope.launch`), while the map stays static. Now:

1. Screen A's loader starts fetching avatar X; the deferred (owned by A's scope) sits in the shared map.
2. Screen B's loader binds the same user and picks up A's deferred from the map.
3. A is destroyed → A's scope cancels → the deferred is cancelled mid-flight → B's `await()` throws `CancellationException` → B's bind coroutine dies and **B's avatar silently never loads**.

That's a realistic sequence (list screen → profile screen navigation) and strictly worse than the status quo. The old design was coherent: static map + static supervisor scope, entries removed in the `finally` after await — requests lived exactly as long as needed, no leak.

**If the goal is killing the static scope, the map must move with it:** either make `inFlightRequests` instance-level too (losing cross-screen dedupe, probably acceptable for 2MB avatars), or keep both static (status quo). Half-migrating ownership is the one option that's wrong.

**Good part:** the `resetForTesting` change (cancel each deferred instead of `cancelChildren`) is fine and would survive either design.

**Cross-PR:** #1082 makes the same half-migration in `DashboardPostImageLoader` and needs the same decision — these two should land together, whichever way they go.

**Verdict:** Coherent-looking diff that breaks shared in-flight dedupe under scope cancellation — needs the map and scope to agree on ownership.

---

dogi:
**Re-review after 825eec4 — fixed. 77/100, lifting changes requested.**

The ownership mismatch is resolved the same way as its twin #1082: `inFlightRequests` is now an instance field (`ConcurrentHashMap`), so shared-map-outliving-owned-deferreds can't happen and the cross-screen cancellation scenario is gone. Removing the in-flight cleanup from the static `resetForTesting()` is the coherent consequence — the map is no longer static state to reset.

Nits (non-blocking): the `bind` path still wraps the now-concurrent map in `synchronized(inFlightRequests)` — harmless (and actually still needed for the check-then-put atomicity, since `getOrPut` on a `ConcurrentHashMap` via the Kotlin extension isn't atomic), but worth a comment so nobody "simplifies" it away; `java.util.concurrent.ConcurrentHashMap` import landed after the `org.ole` group.

**Verdict:** Fixed as prescribed — ready; land together with #1082.

---

PlataformasInformaticas (approved):

PR #1081, opened 2026-07-03
Inject DashboardSurveysRepository in DashboardTeamSurveysFragment

dogi:
🎯 **What:**
Created `DashboardSurveysRepositoryFactory` to act as a provider for `DashboardSurveysRepository` and refactored `DashboardTeamSurveysFragment` to use this factory instead of directly instantiating the repository.

🛡️ **Solution:**
- Created `DashboardSurveysRepositoryFactory.kt` implementing a thread-safe singleton pattern.
- Updated `DashboardTeamSurveysFragment.kt` to inject `DashboardSurveysRepositoryFactory.getRepository()`.

This sets the foundation for easier unit testing by allowing mock repository injection using the factory's `setRepository()` method.

---
*PR created automatically by Jules for task [6035980335084930111](https://jules.google.com/task/6035980335084930111) started by @dogi*

PR #1084, opened 2026-07-03
Refactor: Inject DashboardTeamsRepository in TeamsFragment

dogi:
🎯 **What:** Decoupled `TeamsFragment` from the direct instantiation of `DashboardTeamsRepository` by utilizing a provider factory approach (`DashboardTeamsDependencies`).
🎯 **Why:** To improve code modularity, simplify dependency management, and allow for cleaner, reflection-free unit testing.
🛠️ **How:**
- Created `DashboardTeamsDependencies.kt` to act as a dependency provider with override capabilities for testing.
- Refactored `TeamsFragment` to fetch its repository via `DashboardTeamsDependencies.provideRepository()`.
- Updated `TeamsFragmentTest` to override the repository through the new provider and removed brittle Java reflection code. Added lifecycle hooks to reset the testing state.

---
*PR created automatically by Jules for task [2528368851554434372](https://jules.google.com/task/2528368851554434372) started by @dogi*


dogi:
**Review — 68/100**

**Verified:** The new `DashboardTeamsDependencies` provider works as described, and the test change is the genuinely good part: it deletes the `getDeclaredField("repository")` reflection hack in `TeamsFragmentTest` and replaces it with a proper override + `resetForTesting()` in setup/teardown — that's a real testability improvement, done in the style of the existing `AuthDependencies`.

**Concerns:**
- **Direct collision with #1085:** both PRs create `dashboard/DashboardTeamsDependencies.kt` — same object name, *different* method names (`provideRepository` here vs `provideTeamsRepository` there). Whichever merges second conflicts. #1080 adds a third spelling (`DashboardTeamsRepositoryProvider`) of the same idea. One provider object must win; three cannot coexist.
- **Partial conversion:** only `TeamsFragment` uses the provider; `DashboardTeamMembersFragment` and `DashboardTeamMemberProfileActivity` keep constructing `DashboardTeamsRepository()` directly. A test-override hook that only some call sites respect is a trap — override it in a test that exercises the other screens and nothing happens.
- `TeamsFragment.kt:56` — `get() = provideRepository()` builds a **new repository on every property access** (each with its own... well, after #1094, the shared client, but still new Moshi/Retrofit wiring per access). Cache it in the provider (like #1081's factory does) or keep a `val`.
- Interaction with #1094: that PR (constructor-injects the shared `OkHttpClient`, all 4 call sites, matches project convention) touches the same initializer line. Best end state: **merge #1094 first, rebase this so the provider returns a cached repo built on the shared client, convert all call sites, close #1080 and #1085.**

**Verdict:** The right test-hook idea and the best of the three provider PRs, but needs one round: dedupe against #1085/#1080, cover all call sites, stop re-instantiating per access.

---

dogi:
**Re-review after bb0edfe — fixed. 76/100, lifting changes requested.**

All three substantive flags addressed:
- **Per-access construction gone:** `provideRepository()` now returns a cached instance behind `synchronized` double-check, cleared in `resetForTesting()`. The `TeamsFragment` `get()` property is now cheap.
- **Full conversion:** `DashboardTeamMembersFragment` and `DashboardTeamMemberProfileActivity` now route through the same provider, so the test override governs every screen.
- This makes #1080 and #1085 strict subsets — they should now be closed (their single call sites are both covered here).

Remaining coordination (not a defect): #1094 constructor-injects the shared `OkHttpClient` at these same lines; after it lands, this provider's `DashboardTeamsRepository()` gains a `client = ...` argument — one-line rebase, whichever goes second. Nit: `cachedRepository` isn't `@Volatile` while `repositoryOverride` is; the `synchronized` read path makes it safe in practice, but annotate it for symmetry.

**Verdict:** Flags fixed, now the definitive teams-provider PR — ready; close #1080/#1085.

---

PlataformasInformaticas (approved):

PR #1086, opened 2026-07-03
Refactor `downloadPdf` into `DashboardResourcesRepository`

dogi:
Refactor `downloadPdf` logic into `DashboardResourcesRepository` to keep the activity slim and testable.
Also cleaned up some unnecessary imports and added Robolectric tests to verify the cache download behavior on success and HTTP failure.

---
*PR created automatically by Jules for task [11756535879331937234](https://jules.google.com/task/11756535879331937234) started by @dogi*


dogi:
**Review — 78/100**

**Verified:** Line-for-line faithful move of `downloadPdf`/`saveToTempFile` from the activity into `DashboardResourcesRepository.downloadPdfToCache` — the `file://` short-circuit, https-only `Authorization` header, `runCatching` swallow, and temp-file naming are all preserved; `cacheDir` becomes a parameter, which is the right decoupling. The new tests are real: the success test also pins the subtle "no auth header over plain http" behavior (`assertNull(request.getHeader("Authorization"))`) — good catch to encode that.

**Concerns:**
- `FullscreenPdfActivity.kt:41` — fully-qualified `org.ole.planet.myplanet.lite.dashboard.DashboardResourcesRepository()` inline; add an import like every other reference in the file.
- `DashboardResourcesRepositoryTest.kt:14` — `@RunWith(RobolectricTestRunner::class)` now applies to the *whole existing class*, dragging the previously-plain-JVM tests into Robolectric (slower boot per class) just because `downloadPdfToCache` calls `android.net.Uri.parse`. Consider replacing `Uri.parse(url).scheme == "file"` with `url.startsWith("file:")`/`URI(url)` in the repository so no Android dependency is needed, or split the PDF tests into their own Robolectric class.
- Test style: fully-qualified `org.junit.Assert.assertNotNull(...)` calls instead of imports; missing trailing newline at EOF.
- **Cluster:** `DashboardResourcesRepository.kt` is currently modified by #1070/#1071/#1090/#1101/#1102/#1104/#1131, and `FullscreenPdfActivity` by #1092/#1133 — expect rebases; this one's additions are append-only so it should rebase cleanly.

**Verdict:** Good extraction with genuine tests — mergeable after the cosmetic pass (imports/EOF), ideally dropping the `Uri` dependency to keep the test class JVM-pure.

---

PlataformasInformaticas (approved):

PR #1089, opened 2026-07-03
Refactor: Replace GlobalScope with lifecycleScope in CreateVoiceActivity

dogi:
🎯 **What:**
Replaced `GlobalScope.launch` with `lifecycleScope.launch` in `CreateVoiceActivity`.

⚠️ **Risk:**
Previously, `GlobalScope` was used for file cleanup, which could theoretically outlive the app process if not handled carefully, or at least caused a lint/delicate API warning. The new approach correctly scopes to the activity's lifecycle (which runs upon destruction but is cleanly managed).

🛡️ **Solution:**
- Removed `GlobalScope` imports and usages.
- Integrated `lifecycleScope.launch(Dispatchers.IO)` in `onDestroy` to handle temporary file and bitmap cleanup.
- Ran robolectric tests successfully.

---
*PR created automatically by Jules for task [11109744270724262253](https://jules.google.com/task/11109744270724262253) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/CreateVoiceActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/CreateVoiceActivity.kt
@@ -52,7 +52,6 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.sync.Semaphore
 import kotlinx.coroutines.sync.withPermit
 import kotlinx.coroutines.withContext
@@ -232,8 +231,7 @@ class CreateVoiceActivity : BaseActivity() {
         val filesToDelete = pendingImages.values.map { it.file }.toList()
         pendingImages.clear()
 
-        @OptIn(kotlinx.coroutines.DelicateCoroutinesApi::class)
-        GlobalScope.launch(Dispatchers.IO) {
+        lifecycleScope.launch(Dispatchers.IO) {
             filesToDelete.forEach { file ->
                 if (file.exists()) {
                     file.delete()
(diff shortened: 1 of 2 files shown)


dogi:
**Review — 30/100**

**Verified — this change breaks the cleanup it refactors.** The only call site of this `GlobalScope` block is `CreateVoiceActivity.onDestroy()` (line ~230). `lifecycleScope` is cancelled when the lifecycle hits `ON_DESTROY`, and androidx dispatches that event **before** your `onDestroy()` body runs (`onActivityPreDestroyed` / ReportFragment). Launching on a cancelled scope is a silent no-op — the coroutine is born cancelled and the body never executes. Net effect: pending temp image files are **never deleted again**; every abandoned voice-composer session leaks its images to disk permanently. CI can't catch this — it compiles fine and no test covers it.

**The premise is understandable** — `GlobalScope` is flagged as delicate for good reason — but this is the one situation it (or an equivalent app-level scope) exists for: work that must outlive the component that requests it. The original author even opted in explicitly with `@OptIn(DelicateCoroutinesApi::class)`.

**Fix options, in order of preference:**
1. An application-scoped singleton (`CoroutineScope(SupervisorJob() + Dispatchers.IO)` owned by an object/Application) — exactly what #1126/#1129 introduce (`ApplicationScope.io`). Rebase onto that and use it here.
2. `NonCancellable`/`WorkManager` if you want delivery guarantees.
3. Leave `GlobalScope` — it was correct, just unfashionable.

**Cross-PR:** #1120 contains this identical (equally broken) hunk plus the same change in `DashboardPostDetailActivity`; #1129 changes this same line correctly via `ApplicationScope.io`. **This PR and the CreateVoice half of #1120 are superseded by #1129's approach — recommend closing this one.**

**Verdict:** Swaps working cleanup for a silent no-op — do not merge as-is.

---

dogi:
**Re-review after b49d2cb — fixed. 75/100, lifting changes requested.**

The dead-code cleanup is resolved correctly: `util/ApplicationScope.io` (SupervisorJob + Dispatchers.IO) survives activity teardown, so the `onDestroy` file deletion actually runs again. Notably, your `ApplicationScope.kt` is byte-identical to the one #1126 introduces at the same path — the two PRs will merge cleanly together, and #1129 (root-package variant) and #1079 (`scope`/`Dispatchers.Default` variant) are now the outliers that should align to this file.

One nit: `org.ole.planet.myplanet.lite.util.ApplicationScope.io.launch { ... }` is written fully-qualified inline — add the import.

**Verdict:** The onDestroy cleanup works again — ready; coordinate the single shared `ApplicationScope` with #1126 (identical) and get #1129/#1079 aligned or closed.

---

PR #1092, opened 2026-07-03
⚡ Performance: Render PDF pages asynchronously with caching

dogi:
⚡ **What:** 
Moved PDF page rendering from the main thread to a background coroutine and implemented an LruCache for the rendered bitmaps.

🎯 **Why:**
Rendering PDF pages using `PdfRenderer.render()` is a heavily synchronous operation that blocks the main thread, causing significant UI stutter when scrolling through ViewPager2. Paging backwards previously forced immediate re-rendering of discarded bitmaps, compounding the frame drops.

📊 **Measured Improvement:**
- Rendering is executed safely on `Dispatchers.IO`.
- `LruCache` intelligently stores up to ~12.5% of max available memory worth of bitmaps, completely eliminating stutter when revisiting recent pages.
- Memory safety is enforced by explicitly closing pages, releasing references inside a finally block, and clearing the entire cache when the activity finishes or the views are recycled.

---
*PR created automatically by Jules for task [4029741526449365985](https://jules.google.com/task/4029741526449365985) started by @dogi*


dogi:
**Review — 80/100**

**Verified:** The premise is real — `onBindViewHolder` previously did `openPage`/`render` synchronously on the main thread, which visibly janks page swipes on big PDFs. The async design holds up under scrutiny:
- `renderMutex` correctly serializes access to `PdfRenderer` (not thread-safe) — required, not optional.
- Jobs run in `lifecycleScope`, which is cancelled at `ON_DESTROY` *before* `renderer.close()` runs in the `onDestroy` body — right ordering; and the `catch (e: Exception)` inside the lock covers the residual race where a mid-render job outlives cancellation and hits a closed renderer.
- The `bindingAdapterPosition == position` guard prevents stale binds after recycling; `onViewRecycled` cancels the job. Double-checked-cache inside the lock avoids duplicate renders.
- LruCache sized at maxMemory/8 in KB with `byteCount/1024` sizing is the standard pattern.

**Concerns (nits):**
- `FullscreenPdfActivity.kt:33-34` vs `:198,208` — `Mutex`/`withLock` are imported but then written fully-qualified (`kotlinx.coroutines.sync.Mutex()`); same for `kotlinx.coroutines.CoroutineScope`/`Job`. Pick the imports.
- `catch (e: Exception) → null` silently eats render failures; a `Log.w` would save future debugging.
- Recycled pages show blank until re-render (no placeholder/progress) — acceptable, worth knowing.

**Cross-PR:** #1133 rewrites this same `PdfPageAdapter` to `ListAdapter` and #1086 extracts `downloadPdf` from this activity — this PR conflicts with #1133 directly (both restructure the adapter). This one carries the substantive improvement; **#1133 should rebase on top of (or fold into) this**.

**Verdict:** Real, well-executed perf fix — mergeable after the import cleanup; coordinate with #1133.

---

PlataformasInformaticas (approved):

PR #1085, opened 2026-07-03
Refactor: Inject DashboardTeamsRepository in DashboardTeamMemberProfileActivity

dogi:
🎯 **What:**
Replaced the direct instantiation of `DashboardTeamsRepository` in `DashboardTeamMemberProfileActivity` with dependency injection using a new factory provider.

🛠 **How:**
- Created `DashboardTeamsDependencies` to serve as a manual dependency provider.
- Updated `DashboardTeamMemberProfileActivity` to use `DashboardTeamsDependencies.provideTeamsRepository()` instead of `DashboardTeamsRepository()`.
- The factory object includes methods like `overrideRepository` to make it easier to mock the dependency in future testing.

---
*PR created automatically by Jules for task [6448862740867706374](https://jules.google.com/task/6448862740867706374) started by @dogi*

PR #1094, opened 2026-07-03
Refactor DashboardTeamsRepository to use injected OkHttpClient

dogi:
🎯 **What:**
- Removed the default `OkHttpClient` instantiation from `DashboardTeamsRepository`'s constructor.
- Updated `DashboardTeamMembersFragment`, `TeamsFragment`, and `DashboardTeamMemberProfileActivity` to pass the shared `SharedBitmapDependencies.client`.
- Fixed `DashboardTeamsRepositoryTest` by passing a custom instance explicitly.

🎯 **Why:**
This ensures that `DashboardTeamsRepository` and its callers reuse the same `OkHttpClient` instance (`SharedBitmapDependencies.client`) rather than creating a new builder and instance each time it is instantiated, optimizing network resource utilization and connection pooling.

---
*PR created automatically by Jules for task [15289890074098662388](https://jules.google.com/task/15289890074098662388) started by @dogi*


dogi:
**Review — 84/100**

**Verified:** Removing the defaulted `OkHttpClient` from `DashboardTeamsRepository`'s constructor and wiring the shared client is done completely — I checked every instantiation on main (`TeamsFragment:55`, `DashboardTeamMembersFragment:45`, `DashboardTeamMemberProfileActivity:59`, plus the test) and all four are updated, so nothing is left relying on the deleted default. CI build + test green. This matches the project's documented convention (repositories constructor-injected with `OkHttpClient`/`Moshi`/dispatcher).

**Strengths:**
- Real resource win: three separate `OkHttpClient`s (each with its own connection pool and thread arrangements) collapse into one shared pool.
- Making the client parameter *required* is the right call — it prevents the next caller from silently minting another client.

**Concerns:**
- `SharedBitmapDependencies.client` now serves general team networking, so the name is a lie waiting to confuse someone. Rename it (`SharedHttpDependencies`?) or move `client` somewhere neutral — fine as a follow-up given #1129/#1131 are also leaning on this object.
- `Moshi` keeps its defaulted construction; fine, but the same sharing argument applies if you're in there again.

**Cross-PR — this is the keeper of the teams-repo family:** #1080, #1084 and #1085 each invent a *different* provider object (`DashboardTeamsRepositoryProvider`, two incompatible versions of `DashboardTeamsDependencies`) for single call sites, and all three touch the same initializer lines as this PR. Constructor injection here is the codebase convention; the only thing the providers add is a test-override hook, which #1084 does use well. Recommendation: **merge this**, then rebase #1084's test-hook idea on top (provider returning a shared-client repo) and **close #1080/#1085**.

**Verdict:** Convention-consistent consolidation, complete call-site coverage — mergeable; expect to arbitrate the conflicts with the provider PRs.

---

PlataformasInformaticas (approved):

PR #1098, opened 2026-07-03
Refactor CourseAdapter to ListAdapter

dogi:
Refactored `CourseAdapter` in `DashboardCoursePageFragment` to extend `ListAdapter`. 
- Introduced a sealed class `CourseListItem` to allow seamless diffing between the header and course items. 
- Consolidated all `update*` methods to apply their changes and recompute the final list to call `submitList`, avoiding error-prone manual index bookkeeping.
- Optimized download progress updates by emitting a "PROGRESS_UPDATE" payload via `DiffUtil.ItemCallback` and handling partial binding via `onBindViewHolder(payloads)`, reducing unnecessary view recreations.

---
*PR created automatically by Jules for task [12578292787324861392](https://jules.google.com/task/12578292787324861392) started by @dogi*


dogi:
**Review — 74/100** · **best of the four CourseAdapter rewrites (#1098/#1127/#1140/#1145) — keep this one, with one fix**

**Verified:** This is the only member of the cluster that fully adopts `ListAdapter` semantics: download state (`isDownloaded`, `downloadProgress`) moves *into* the diffed item (`CourseListItem.CourseItemWrapper`), so every mutation funnels through `applyFilter()`/`submitList` and DiffUtil computes minimal updates — no leftover manual `notifyItem*` calls for row data. The `"PROGRESS_UPDATE"` change payload plus the `onBindViewHolder(payloads)` override correctly avoids re-binding images on every progress tick. Header/wrapper diffing is sound (`data object Header`, id-based identity).

**Two behavior changes to be aware of (both arguably fixes, neither in the description):**
1. `submitCourses` now routes through `applyFilter()`, so an active search query/tag filter survives a data refresh — the old code reset to the unfiltered list. Better behavior, but say so.
2. **The header is no longer guaranteed to exist.** Old `getItemCount() = displayedItems.size + 1` meant the header (search box + categories) rendered from frame one; now the list is empty until the first `applyFilter()`. Worse: `updateCategories()` (line 1267) still calls `notifyItemChanged(0)` — a manual notify against a possibly-empty `ListAdapter` (categories load at fragment line 302 can race the first `submitCourses`). Fix both with one line: `init { submitList(listOf(CourseListItem.Header)) }` — and ideally replace the `notifyItemChanged(0)` with a resubmit so no manual notifies remain.

**Cross-PR:** #1127 and #1140 are `ListAdapter<Any>`-with-sentinel versions that keep manual notifies (half-migrations); #1145 keeps `RecyclerView.Adapter` + `AsyncListDiffer` with offset callbacks (faithful but least ambitious). **Recommend: land this one after the init fix; close #1127/#1140/#1145.** Note #1105/#1124/#1132 make small targeted edits to this same fragment and will need rebasing on whichever rewrite lands.

**Verdict:** The right architecture, one init-ordering hole — one round.

---

dogi:
**Re-review after latest push — not fixed; keeping changes requested (and this one's on me for under-specifying).**

The `init { submitList(listOf(CourseListItem.Header)) }` half is right — the header now exists from frame one. But replacing `notifyItemChanged(0)` with `applyFilter()` in `updateCategories()` doesn't do what we both hoped: `applyFilter` resubmits a list whose header element is the singleton `data object Header`, so the diff callback sees `Header == Header`, reports no change, and **never rebinds the header** — meaning newly loaded categories still never render chips. My earlier suggestion to "replace the notify with a resubmit" missed that a constant object can't carry change information; apologies for pointing you at half a fix.

**The actual fix:** make the header item carry the state its ViewHolder renders — e.g. `data class Header(val categories: List<CourseCategory>, val selectedIndex: Int, val searchQuery: String) : CourseListItem()` — and build it in `applyFilter()`. Then category changes diff as content changes and the rebind happens exactly when needed, with zero manual notifies. (Alternatively keep the plain `notifyItemChanged(0)` in `updateCategories()` — unfashionable but correct.)

**Verdict:** Init hole fixed, but category updates are now silently dropped — one more round with the stateful Header item.

---

dogi:

The stateful-Header fix landed exactly as discussed: `Header` is now a `data class` carrying categories/selectedIndex/searchQuery, built fresh in `applyFilter()`; the ViewHolder binds from the item; the init block submits a populated header; and the chip-selection callback routes through `applyFilter()` — zero manual notifies remain, and category loads now diff as content changes and rebind correctly. Nit: the header now rebinds on every search keystroke (searchQuery is in the item), so `setText` in bind can jump the cursor on mid-string edits — consider skipping `setText` when the text is unchanged.

**Verdict:** flag fixed — ready; still the CourseAdapter cluster keeper (close #1127/#1140/#1145, rebase #1105/#1124/#1132/#1150 remnants).

---

PlataformasInformaticas (approved):

PR #1099, opened 2026-07-03
Refactor: Constructor-inject Coroutine dispatchers in dashboard stores

dogi:
This PR refactors `DashboardSurveyOutboxStore` and `DashboardOfflineSurveyStore` to inject Coroutine dispatchers via their constructors.

* Adds `ioDispatcher` (and `defaultDispatcher` where necessary) to the constructors of both stores.
* Replaces inline usages of `Dispatchers.IO` and `Dispatchers.Default` with the injected dispatchers.
* Updates local and instrumentation tests to instantiate the stores with `UnconfinedTestDispatcher` and removes real-IO waits like `delay(10)`.

---
*PR created automatically by Jules for task [11024506999130051594](https://jules.google.com/task/11024506999130051594) started by @dogi*

### DIFF
--- a/app/src/androidTest/java/org/ole/planet/myplanet/lite/dashboard/DashboardOfflineSurveyStoreTest.kt
+++ b/app/src/androidTest/java/org/ole/planet/myplanet/lite/dashboard/DashboardOfflineSurveyStoreTest.kt
@@ -6,6 +6,9 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.squareup.moshi.Moshi
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runTest
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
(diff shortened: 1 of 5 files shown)


dogi:
**Review — 80/100**

**Verified:** Complete and correct — every `withContext(Dispatchers.IO)` / `withContext(Dispatchers.Default)` in both stores now goes through injected `ioDispatcher`/`defaultDispatcher` params defaulting to the originals, so production behavior is unchanged (`getInstance` still constructs with defaults). Unlike some siblings, the tests actually *use* the new seams (`UnconfinedTestDispatcher` in both the unit and instrumented suites) and migrate `runBlocking` → `runTest`. The positional-argument hazard of inserting `ioDispatcher` before `moshi` is handled — the one direct `(context, moshi)` caller (androidTest) was switched to named arguments. CI green.

**Strengths:**
- Distinguishing `ioDispatcher` (DB) from `defaultDispatcher` (JSON parsing) preserves the original threading intent instead of collapsing them.
- `private constructor` → `internal constructor` on the outbox store is the minimal visibility bump needed for the test seam; the singleton contract survives for production callers.

**Concerns (nits):**
- Unit tests now construct stores directly instead of `getInstance`, creating a new `SQLiteOpenHelper` per test against the same db name — works under Robolectric, but two live helpers on one file is the kind of thing that surprises later; a `resetForTesting()` on the singleton would be tidier.
- New imports appended out of sorted order in all four test files.

**Verdict:** Convention-consistent, test-backed dispatcher injection done properly — mergeable.

---

PlataformasInformaticas:

PR #1106, opened 2026-07-03
Refactor survey completion count mapping logic to repository

dogi:
Added a repository helper `getCompletionCountsWithDefaults` that retrieves completion counts for surveys and populates zero defaults for ids missing from the response. This simplifies the fragment logic in `DashboardTeamSurveysFragment.kt` and removes the need for direct Coroutines `withContext(Dispatchers.IO)` and manual tracking manipulation from the caller side. Covered the new helper method with unit tests in `DashboardSurveysRepositoryTest.kt`.

---
*PR created automatically by Jules for task [749707321679995703](https://jules.google.com/task/749707321679995703) started by @dogi*

PR #1107, opened 2026-07-03
Refactor Dashboard surveys load into repository state class

dogi:
The goal of this change is to extract the complex offline, network, and caching data loading orchestration in the `DashboardTeamSurveysFragment` into the repository tier.

I accomplished this by:
1. Creating a `SurveysLoadState` data class in the `DashboardSurveysRepository` that maps the returned data fields from the view (`teamSurveys`, `adoptedSurveys`, `completionCounts`, `savedSurveyIds`, `savedSurveyRevisions`, `outboxEntries`).
2. Moving the core orchestration logic out of the `DashboardTeamSurveysFragment`'s `loadSurveys` method and into a newly defined `loadSurveysState()` method inside the `DashboardSurveysRepository`. This method properly cascades data logic and determines counts by passing in the `DashboardLocalSurveyRepository`.
3. Updating the fragment to handle UI actions and error toasts while feeding the returned result model cleanly into its adapter bindings, removing the previous `fetchCompletionCounts()` method completely.

---
*PR created automatically by Jules for task [16655284187346632625](https://jules.google.com/task/16655284187346632625) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamSurveysFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamSurveysFragment.kt
@@ -165,6 +165,7 @@ class DashboardTeamSurveysFragment : Fragment(R.layout.fragment_dashboard_team_s
         }
     }
 
+
     private fun loadSurveys(isSwipeRefresh: Boolean = false) {
         val base = baseUrl
         val team = teamId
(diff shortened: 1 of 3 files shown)


dogi:
**Review — 66/100**

**Verified:** The orchestration move is faithful where it matters — I checked the tricky paths: offline mode reads local only; online fetch failure falls back to cache and errors only when the cache is empty (preserved, now via a wrapped `IOException`); completion counts keep the zeros-then-merge semantics; saved ids/revisions/outbox loads unchanged. The `SurveysLoadState` value object is a reasonable shape. The repository's new null checks on base/team/credentials are unreachable in practice (the fragment's specific-message guards at `loadSurveys` top remain), so no user-facing change there — just redundant validation with worse messages if anyone ever hits it.

**Concerns:**
- **No tests.** This is the largest logic move in the surveys area — multi-branch fallback behavior now sits in a `suspend` repository method that MockWebServer + a fake local repository could test directly, and nothing was added. The offline-fallback-when-cache-empty branch in particular has never been pinned.
- **Layering:** `DashboardSurveysRepository` now imports `survey.DashboardLocalSurveyRepository` and `DashboardSurveyOutboxStore.OutboxEntry` and takes the local repository as a method parameter — it's become a use-case/interactor wearing a repository's name. Acceptable if the team wants that pattern, but then `localRepository` belongs in the constructor, not threaded through each call.
- **Duplicates #1106:** the counts-with-defaults block here re-implements what #1106 extracts and tests. Merge #1106 first and call `getCompletionCountsWithDefaults` from here.
- Stray double blank lines (`fragment:168`, repo `:54`, `:390`); import order.

**Verdict:** Sound consolidation with the fallback semantics intact, but a move this size without a single test isn't done — one round (tests + reuse #1106 + constructor injection of the local repo).

---

dogi:
**Re-review after latest push — main flag fixed. 75/100, lifting changes requested.**

The blocking issue (zero tests on the largest surveys-area logic move) is resolved: `loadSurveysState` now has ~100 lines of tests covering offline mode with a mocked local repository (team/adopted split, zeroed completion counts, saved ids/revisions/outbox propagation) — the branchy behavior is finally pinned. Stray blank lines cleaned up too.

Not taken (acceptable as follow-ups, no longer blocking): the completion-counts block still re-implements what #1106 extracts (if #1106 merges first, swap the inline block for `getCompletionCountsWithDefaults` — two lines), and `localRepository` remains a method parameter rather than a constructor dependency.

**Verdict:** Tested now — ready; sequence after #1106 to dedupe the counts logic.

---

PlataformasInformaticas:

PR #1108, opened 2026-07-03
♻️ Extract Profile and Credential Lookups into ProfileCredentialsStore Accessor

dogi:
🎯 **What:**
- Added `CurrentUserSession` data class to `ProfileCredentialsStore` to hold both `StoredCredentials` and `UserProfile`.
- Created `getCurrentUserSession` in `ProfileCredentialsStore` to fetch the session info via an injected CoroutineDispatcher (defaults to IO), pulling from `SecurePreferences` and `UserProfileDatabase`.
- Refactored `DashboardPostDetailActivity`'s `initializeSession` to fetch this composite object and cache the credentials and profile locally.
- Updated scattered validation logic in `attemptPostReply`, `attemptSaveCommentEdit`, `loadCachedProfile`, and `buildUserPayload` to use the cached local copies rather than hitting the DB or Preferences.
- Removed unused `UserProfileDatabase` imports from the activity.

📊 **Coverage:**
- Verified `DashboardPostDetailActivityRobolectricTest` continues to pass, ensuring mocked credentials and profile scenarios behave as expected.
- Verified `ProfileCredentialsStoreTest` passes correctly.

✨ **Result:**
Reduced redundant local I/O in the `DashboardPostDetailActivity` lifecycle, centralizing user-context access into a single cached provider point.

---
*PR created automatically by Jules for task [7840973079901136430](https://jules.google.com/task/7840973079901136430) started by @dogi*


dogi:
**Review — 68/100**

**Verified:** The consolidation works and removes real duplication (two separate `UserProfileDatabase.getInstance(...).getProfile()` reads and repeated `getStoredCredentials` calls collapse into one `getCurrentUserSession` at `initializeSession`). The new store method is small and has an injectable dispatcher.

**Concerns — the caching semantics quietly changed:**
- `loadCachedProfile()` (`DashboardPostDetailActivity.kt:1278`) used to be *lazy with retry*: if `cachedProfile` was null it re-read the database on every call until a profile appeared. Now it just returns the snapshot taken at `initializeSession`. If the profile row lands in the DB after the activity starts — exactly what `UserProfileSync` does asynchronously after login — this activity now sees `null` forever. Same for `buildUserPayload` (`:1305`): reply payloads (names, metadata) built from a possibly-stale/absent snapshot where they previously read fresh. This isn't hypothetical churn; it changes what gets sent in comment payloads in the sync-raced window. Either keep the lazy fallback (`cachedProfile ?: fetchAndCache()`) or state the snapshot semantics as intended in the description.
- `getCurrentUserSession` returns `null` (and thus no profile) when credentials are absent — coupling profile availability to credential availability, which the old code didn't do. Probably harmless here, but it's a semantic narrowing hiding in a helper.
- No test for `getCurrentUserSession` despite the injectable dispatcher existing solely for that purpose.
- **Cluster:** `DashboardPostDetailActivity` is contested (#1079/#1110/#1120/#1121/#1126/#1142) — rebases likely.

**Verdict:** Good dedup with an unacknowledged behavioral narrowing — restore the lazy fallback (or justify the snapshot) and add the store test; one round.

---

dogi:
**Re-review after latest push — the secondary flag was fixed, the primary one wasn't; keeping changes requested.**

What improved: `getCurrentUserSession` now has three real tests (no credentials → null; credentials + missing profile → session with null profile; both present) using the injectable dispatcher. Good.

What's unchanged — and was the blocking concern: `loadCachedProfile()` (`DashboardPostDetailActivity.kt:1278`) still just `return cachedProfile`, and `buildUserPayload` still reads the snapshot. The lazy-retry semantics of the original code (re-read the DB until a profile appears) are still gone, so a profile that lands via `UserProfileSync` *after* this activity starts is never picked up, and reply payloads in that window are built without names. The fix is one line: `return cachedProfile ?: withContext(Dispatchers.IO) { UserProfileDatabase.getInstance(applicationContext).getProfile() }.also { cachedProfile = it }` — or an explicit statement in the PR description that snapshot semantics are intended and why.

**Verdict:** Tests good, regression still in place — one more round.

---

dogi:

The snapshot regression is resolved with the exact one-liner: `loadCachedProfile()` is lazy-with-retry again (`cachedProfile ?: DB read .also { cachedProfile = it }`), and `buildUserPayload` routes through it. Profiles landing via `UserProfileSync` after activity start are picked up again. With the store tests from the previous push, all flags closed.

**Verdict:** fixed — ready.

---

PlataformasInformaticas (approved):

PR #1110, opened 2026-07-03
Refactor image cleanup to use lifecycleScope instead of GlobalScope

dogi:
Replaced `GlobalScope.launch` with `lifecycleScope.launch` for image cleanup in `DashboardPostDetailActivity`. The coroutine `Job` is tracked and explicitly cancelled early in `onDestroy()` before other teardown operations run.

---
*PR created automatically by Jules for task [3627743076070406597](https://jules.google.com/task/3627743076070406597) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
@@ -53,7 +53,6 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import org.ole.planet.myplanet.lite.R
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 30/100**

**Verified — this makes cleanup worse in two ways, not better.** `clearPendingReplyImages()` is called both mid-lifecycle (after posting/cancelling a reply — fine for `lifecycleScope`) and from **`onDestroy()` (line 1309)**. Two problems:

1. **The onDestroy launch is a no-op.** `lifecycleScope` is cancelled when `ON_DESTROY` dispatches, which happens *before* your `onDestroy()` body runs — so the cleanup coroutine launched from the destroy path is born cancelled and the pending temp files are never deleted. This is the same flaw as #1079/#1089/#1120.
2. **The new `imageCleanupJob?.cancel()` (line 1301) actively kills legitimate cleanup.** If the user cleared their reply images moments before leaving the screen, that deletion may still be running on IO — and this line cancels it on destroy. The old `GlobalScope` code guaranteed those files got deleted; this PR guarantees the opposite. Tracking the job only to cancel it is backwards for fire-and-forget deletion — there's nothing to "leak" here that outlives a few file deletes.

**The correct pattern for this site** is an application-level scope: #1126 converts this exact block to `ApplicationScope.io` (new `util/ApplicationScope`), which survives activity teardown and needs no job juggling. 

**Cross-PR:** #1079 and #1120 modify this same block (both with the same onDestroy flaw, without the added cancel); #1126 does it correctly. **Recommend closing this in favor of #1126.**

**Verdict:** Converts working cleanup into a no-op and adds a cancel that kills in-flight deletions — do not merge.

---

PR #1115, opened 2026-07-03
feat: enrich join requests in repository

dogi:
Moved the user profile lookup and merging logic for join requests from DashboardTeamMembersFragment into DashboardTeamsRepository to encapsulate data assembly. The fragment now cleanly consumes a list of pre-assembled TeamJoinRequestUiModel objects.

---
*PR created automatically by Jules for task [15870863711445644226](https://jules.google.com/task/15870863711445644226) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamMembersFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamMembersFragment.kt
@@ -254,9 +254,9 @@ class DashboardTeamMembersFragment : Fragment() {
 
         val teamPlanetCodeForRequests = currentTeamPlanetCode ?: serverPlanetCode
         val joinRequests = if (teamPlanetCodeForRequests.isNullOrBlank()) {
-            emptyList<JoinRequestDocument>()
+            emptyList<TeamJoinRequestUiModel>()
         } else {
-            repository.fetchTeamJoinRequests(
+            repository.fetchEnrichedTeamJoinRequests(
                 baseUrl = base,
                 credentials = creds,
                 sessionCookie = sessionCookie,
@@ -268,7 +268,7 @@ class DashboardTeamMembersFragment : Fragment() {
             }
         }
         if (joinRequests != null) {
-            loadJoinRequestsData(base, creds, joinRequests)
+            loadJoinRequestsData(joinRequests)
         } else {
             hideJoinRequestsSection()
         }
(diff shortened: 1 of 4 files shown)


dogi:
**Review — 60/100**

**Verified:** The enrichment logic itself moved faithfully — username fallback from the CouchDB id, full-name assembly, avatar flag, case-insensitive sort are identical. But the failure semantics changed:

- **Profile-fetch failure now hides the whole section.** Old fragment code: `repository.fetchUserProfiles(...).getOrDefault(emptyList())` — if profiles couldn't load, every request still rendered with the username fallback. New code calls `operations.fetchUserProfiles(...)` *inside* the enrichment's `runCatching`, so a thrown profile-fetch error fails the entire `fetchEnrichedTeamJoinRequests`, the fragment gets a failed `Result`, and `hideJoinRequestsSection()` runs — join requests silently vanish on a flaky profiles call that used to degrade gracefully. Wrap the profiles call in its own `runCatching { ... }.getOrDefault(emptyList())` inside the repository to restore the old behavior.
- **Layering:** `DashboardTeamsRepository` now imports and returns `TeamJoinRequestUiModel` from the root UI package (`DashboardTeamsRepository.kt:16`). A repository returning a *UiModel* inverts the dependency direction the rest of this refactor campaign is establishing. Either move the model into `dashboard/` as a plain domain type (`EnrichedJoinRequest`) or keep the UI mapping in the fragment.
- No test, though the enrichment (name assembly, fallback, sorting) is exactly the pure-ish logic that's now one MockWebServer/fake away from being testable.

**Cross-PR:** touches `DashboardTeamMembersFragment` (also #1080/#1094/#1123) and `DashboardTeamsRepository` (also #1094) — rebase after #1094.

**Verdict:** Right direction, but it swaps graceful degradation for all-or-nothing and leaks a UI type into the repository — needs a round.

---

dogi:
**Re-review after latest push — fixed. 74/100, lifting changes requested.**

Both flags addressed for real:
- **Graceful degradation restored:** the profiles fetch inside `fetchEnrichedTeamJoinRequests` is now wrapped in `runCatching { ... }.getOrDefault(emptyList())`, so a flaky profiles call degrades to username-only rows again instead of hiding the whole join-requests section.
- **Layering fixed:** the repository returns a `dashboard.EnrichedJoinRequest` domain type; the `TeamJoinRequestUiModel` mapping moved back into the fragment.

Nit: the fragment writes `org.ole.planet.myplanet.lite.dashboard.EnrichedJoinRequest` fully-qualified twice — import it. Still no test on the enrichment, which is now trivially fakeable — worth a follow-up.

**Verdict:** Both regressions resolved — ready.

---

PlataformasInformaticas:

PR #1119, opened 2026-07-03
Refactor: remove redundant nested Dispatchers.IO in DashboardCoursesRepository

dogi:
This PR optimizes `DashboardCoursesRepository` by removing redundant nested `withContext(Dispatchers.IO)` blocks inside the course progress fetch methods (`fetchCoursesProgress` and `fetchCoursesProgressDocuments`). Because the entire method is already wrapped in `withContext(dispatcher)` (which defaults to `Dispatchers.IO`), the inner context switch is unnecessary and adds minor overhead. The network calls now execute directly on the injected repository dispatcher. All relevant repository unit tests pass successfully.

---
*PR created automatically by Jules for task [12514807323475253387](https://jules.google.com/task/12514807323475253387) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardCoursesRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardCoursesRepository.kt
@@ -353,17 +353,15 @@ class DashboardCoursesRepository(
                     .addHeader("Authorization", Credentials.basic(credentials.username, credentials.password))
                     .build()
 
-                val docs = withContext(Dispatchers.IO) {
-                    client.newCall(request).execute().use { response ->
-                        if (!response.isSuccessful) {
-                            response.body.string()
-                            throw IOException("Unexpected response ${response.code}")
-                        }
-                        val parsed = coursesProgressResponseAdapter.fromJson(response.body.string())
-                            ?: throw IOException("Invalid response body")
-                        parsed.docs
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 85/100**

**Verified:** Both `fetchCoursesProgress` and the stepNum variant already wrap their bodies in `withContext(dispatcher)` at the method top (`DashboardCoursesRepository.kt:329, :386`), so the inner `withContext(Dispatchers.IO)` around the `execute()` blocks was pure redundancy — and worse than redundancy: because it hardcoded `Dispatchers.IO`, it **escaped the injected dispatcher**, meaning a test that injects a `TestDispatcher` still ran the network call on real IO. Removing it is both a simplification and a genuine fix for the class's dispatcher-injection contract. No behavior change in production (outer dispatcher defaults to `Dispatchers.IO`). CI green.

**Strengths:**
- Exactly the right size: deletes the redundancy, touches nothing else, no reformatting noise beyond the necessary dedent.

**Concerns (minor):**
- A one-line test asserting these methods respect an injected dispatcher would pin the fixed contract, but the existing MockWebServer tests already exercise the paths.
- `DashboardCoursesRepository` is also touched by #1088 (adds download methods) — trivial rebase whichever lands second.

**Verdict:** Small, correct, and it quietly fixes a real test-dispatcher leak — merge.

---

PlataformasInformaticas:

PR #1100, opened 2026-07-03
Refactor: Remove GlobalScope pending-image cleanup from CreateVoiceActivity

dogi:
Removed `GlobalScope` pending-image cleanup from `CreateVoiceActivity`. 

This patch introduces a clean `deleteFiles` suspend helper leveraging `withContext(Dispatcher)` and substitutes the delicate `GlobalScope` with a `lifecycleScope` bound to a `NonCancellable` context during `onDestroy`. The implementation precisely addresses the requirements for safe, structured concurrency file deletions. Additionally, a unit test using `StandardTestDispatcher` provides robust coverage for the newly decoupled deletion mechanism.

The analogous block within `DashboardPostDetailActivity` remains untouched as instructed, prepared for a future, simplified integration.

---
*PR created automatically by Jules for task [10816930877769249854](https://jules.google.com/task/10816930877769249854) started by @dogi*

PR #1120, opened 2026-07-03
Refactor: Move cache cleanup off GlobalScope in Voice Activities

dogi:
Replaced `GlobalScope.launch` with `lifecycleScope.launch` in `CreateVoiceActivity` and `DashboardPostDetailActivity` for pending image deletion to ensure proper lifecycle management and avoid unbounded background operations. Tests successfully passed.

---
*PR created automatically by Jules for task [3741019081726154316](https://jules.google.com/task/3741019081726154316) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/CreateVoiceActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/CreateVoiceActivity.kt
@@ -52,7 +52,6 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.sync.Semaphore
 import kotlinx.coroutines.sync.withPermit
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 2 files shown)


dogi:
**Review — 28/100**

**Verified — both halves of this PR break the cleanup they refactor.** This PR is the exact union of #1089 (CreateVoiceActivity hunk, byte-identical) and #1079 (DashboardPostDetailActivity hunk, byte-identical), and inherits the fatal flaw in each:

- `CreateVoiceActivity.kt:234` — the block runs in **`onDestroy()`**. `lifecycleScope` is cancelled when `ON_DESTROY` dispatches, which happens *before* the `onDestroy()` body executes, so `lifecycleScope.launch(Dispatchers.IO)` here is a coroutine born cancelled: the body never runs and pending voice-image temp files are never deleted again. This is the *only* call site, so the PR converts working cleanup into dead code.
- `DashboardPostDetailActivity.kt:778` — `clearPendingReplyImages()` is called from four mid-lifecycle sites (fine) **and from `onDestroy()` (line 1309)** — the destroy-path deletion silently no-ops for the same reason, leaking reply-image temp files on the most common exit route.

The deleted `GlobalScope` + `@OptIn(DelicateCoroutinesApi::class)` was doing exactly what it's for: work that must outlive the component. The idiomatic replacement is an app-level scope — #1126 (`util/ApplicationScope`) and #1129 (`ApplicationScope`) both introduce one and convert these same sites correctly; #1100's `NonCancellable` variant also works (inelegantly).

**Cross-PR:** supersedes #1079 and #1089 (identical hunks — those two should close as subsets regardless), but is itself superseded by the app-scope approach. **Recommend closing all three of #1079/#1089/#1120 in favor of #1126 + #1129.**

**Verdict:** Two silent cleanup-killers in one PR — do not merge.

---

PR #1122, opened 2026-07-03
feat: Debounce invite-member dialog search reloads

dogi:
Added debounce functionality to the invite members search input to prevent firing rapid API calls for every keypress. The query is now delayed by 300ms, and if a new keystroke occurs during that window, the previous job is cancelled. Also added a debounce test.

---
*PR created automatically by Jules for task [18212794531348114831](https://jules.google.com/task/18212794531348114831) started by @dogi*


dogi:
**Review — 70/100**

**Verified:** The debounce works for the common case and — credit where due — this is the only debounce PR in the set that both cancels the pending job on dialog dismiss *and* ships a test that actually advances virtual time to prove one fetch fires for rapid typing.

**Concern — the guard races the assignment:**
- `DashboardTeamMembersSupport.kt:263-269` — `currentSearchTerm = newQuery` was moved *inside* the launched coroutine, but the `if (newQuery != currentSearchTerm)` guard runs synchronously in the listener. Until the job actually executes, `currentSearchTerm` is stale. Sequence: type `"a"` (job scheduled, term still `""`), quickly delete it → listener sees `"" != ""` → **false** → no new job, no cancel — and the pending `"a"` job fires 300ms later, loading results for a query the box no longer contains. Fix is exactly what the twin PR #1147 does: assign `currentSearchTerm = newQuery` synchronously in the listener, then launch the delayed `loadInvitePage`.

**Cross-PR — duplicate:** #1147 debounces this same listener (synchronous assignment — correct — but no dismiss-cancel and no test). Best outcome: **keep this PR, adopt #1147's assignment ordering (one-line move), close #1147.** Related but not duplicate: #1123 debounces the *local* member filter in the fragment.

**Verdict:** The strongest of the two invite-debounce PRs thanks to the test and cleanup — one-line fix to the guard race and it's mergeable.

---

dogi:

Exactly the one-line fix: `currentSearchTerm = newQuery` now happens synchronously in the listener before the cancel/launch, so the `newQuery != currentSearchTerm` guard can never compare against a stale value — the type-and-revert orphaned-search race is closed. Combined with the dismiss-cancel and the virtual-time debounce test this PR already had, it's now the definitive invite-search debounce.

**Cross-PR:** #1147 (the twin that had this ordering right but nothing else) should now close in favor of this.

**Verdict:** Fixed — ready; close #1147.

---

PlataformasInformaticas (approved):

PR #1124, opened 2026-07-03
Narrow course image-loader adapter invalidation

dogi:
- Added an adapter helper that notifies only visible course item positions/rows with image URLs when the loader becomes ready.
- Replaced the broad `notifyDataSetChanged` call in `ensureCourseImageLoader` with this new targeted helper to prevent unnecessary layout invalidations.
- Kept the current `RecyclerView.Adapter` implementation per requirements (did not migrate to `ListAdapter`).

---
*PR created automatically by Jules for task [6043968083098031047](https://jules.google.com/task/6043968083098031047) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
@@ -279,7 +279,7 @@ class DashboardCoursePageFragment : Fragment(R.layout.fragment_dashboard_courses
             val sessionCookie = withContext(Dispatchers.IO) { authService.getStoredToken() }
             courseImageLoader = DashboardPostImageLoader(base, sessionCookie, viewLifecycleOwner.lifecycleScope)
             isCourseImageLoaderLoading = false
-            adapter.notifyDataSetChanged()
+            adapter.notifyImageLoaderReady()
         }
     }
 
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 80/100**

**Verified:** Correct and genuinely narrower — the image loader becoming ready only affects rows that render a cover (`coverPath`), so rebinding exactly those items (offset +1 for the header) preserves behavior while skipping the header rebind and the no-cover rows that `notifyDataSetChanged()` used to blow away, along with its full view-holder invalidation and scroll-state churn. Items without covers bind a placeholder unaffected by the loader, so skipping them is safe.

**Concerns (minor):**
- A run of adjacent cover rows produces N single `notifyItemChanged` ops; RecyclerView coalesces them per layout pass, so this is fine — just noting `notifyItemRangeChanged(1, displayedItems.size)` was the simpler 90% version if the selective loop ever feels fussy.
- **Cluster:** every CourseAdapter→ListAdapter rewrite (#1098/#1127/#1140/#1145) touches this same call site (`displayedItems` ceases to exist in all of them; #1140 even includes a blunter version of this exact change). If #1098 lands as recommended, this helper needs a two-line rework against `currentList`. Merge order matters more than content here.

**Verdict:** Small, correct, measurable improvement — mergeable; coordinate with the adapter-rewrite cluster to avoid churn.

---

PlataformasInformaticas:

PR #1116, opened 2026-07-03
Refactor ResourceExplorerAdapter to ListAdapter

dogi:
Convert ResourceExplorerAdapter to ListAdapter using DiffUtils.itemCallback and payload-aware updates to prevent full row rebinds on targeted progress updates.

---
*PR created automatically by Jules for task [16972296284840259861](https://jules.google.com/task/16972296284840259861) started by @dogi*

PR #1113, opened 2026-07-03
Refactor ResourceExplorerAdapter to ListAdapter

dogi:
Converts the custom RecyclerView.Adapter `ResourceExplorerAdapter` to standard `ListAdapter` backed by `DiffUtils.itemCallback`. The download progress state remains externally managed to allow explicit updates (`notifyItemChanged()`) without requiring full list diff recalculations. All initial and replace list usages in the call sites have been updated to `submitList`.

---
*PR created automatically by Jules for task [4246740652844429396](https://jules.google.com/task/4246740652844429396) started by @dogi*

PR #1125, opened 2026-07-03
Migrate ResourceExplorerAdapter to ListAdapter

dogi:
Migrates the `ResourceExplorerAdapter` to use Android's `ListAdapter` for automated `DiffUtil` calculations instead of manually managing list states and calling `notifyItemRange*` methods. 

This change reduces state-management bugs, safely tracks updates using `currentList`, and leaves `notifyResourceChanged` targeted exclusively for download progress updates. Also introduces testing to verify list behaviors.

---
*PR created automatically by Jules for task [8820679791504026621](https://jules.google.com/task/8820679791504026621) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardResourcesListLoadingExtensions.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardResourcesListLoadingExtensions.kt
@@ -55,9 +55,9 @@ private fun DashboardResourcesPageFragment.refreshTeamResources(
     }
     val currentAdapter = list.adapter as? DashboardResourcesPageFragment.ResourceExplorerAdapter
     if (currentAdapter == null) {
-        list.adapter = DashboardResourcesPageFragment.ResourceExplorerAdapter(teamResourcesItems.toList(), resourceDownloadProgress, ::openResource, ::onSecondaryAction)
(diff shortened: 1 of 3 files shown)


dogi:
**Review — 68/100**

**Verified:** Correct `ListAdapter` migration — constructor drops `initialResources`, all extension call sites move to `submitList(...toList())`, `notifyResourceChanged` centralizes the two copy-pasted progress-update blocks in the fragment. And it's the only member of the five-PR cluster that ships a test.

**Concerns:**
- **The test's second half is vacuous.** In `ResourceExplorerAdapterTest`, the first `submitList` (null→list) commits synchronously so its callback assertions run; but the second `submitList` diffs on a background executor and posts the commit callback to the (paused) Robolectric main looper — the callback, and therefore the assertions inside it, **never execute**. The test passes today for the wrong reason. Fix: `shadowOf(Looper.getMainLooper()).idle()` after submit (plus latch/await on the background diff), then assert outside the callback.
- Dropping the explicit `areContentsTheSame` falls back to `==` — same semantics, fine — but unlike #1116 there are no change payloads, so every progress tick still full-rebinds with the change animation.
- Stray double blank lines around `notifyResourceChanged` (`DashboardResourcesPageFragment.kt:241-243`); fully-qualified `DiffUtils` in the supertype.

**Cluster:** five PRs migrate this adapter (#1113/#1116/#1125/#1130/#1136). #1116 achieves the same with zero call-site churn *plus* payload-based flicker suppression — I've recommended it as keeper, with this PR's test (repaired) ported onto it. **Recommend closing this one after moving the test.**

**Verdict:** Solid migration, best intentions on testing, but the test needs to actually assert — superseded by #1116 + ported test.

---

PR #1095, opened 2026-07-03
⚡ Refactor DashboardNewsActionsRepository to inject shared dependencies

dogi:
🎯 **What:** 
- Exposed `client` and `moshi` variables in `AuthDependencies`.
- Removed default instances of `OkHttpClient`, `Moshi`, and `CoroutineDispatcher` from `DashboardNewsActionsRepository`.
- Updated instantiations of `DashboardNewsActionsRepository` in `CreateVoiceActivity`, `DashboardPostDetailActivity`, `DashboardVoicesFragment`, and tests to inject the shared dependencies from `AuthDependencies`.

⚠️ **Risk:** 
- Low. This is a pure refactoring of dependency instantiation that leverages existing singletons safely.

🛡️ **Solution:** 
- Updated `AuthDependencies.kt` to expose the client and moshi cleanly using existing double-checked locked caches.
- Refactored `DashboardNewsActionsRepository.kt` constructor to strictly require all dependencies.
- Added necessary imports and injected dependencies at all relevant instantiation points across the `lite/dashboard` package.

---
*PR created automatically by Jules for task [8137749156823560207](https://jules.google.com/task/8137749156823560207) started by @dogi*

PR #1126, opened 2026-07-03
Refactor DashboardPostDetailActivity to reuse OkHttpClient and drop GlobalScope

dogi:
Refactored `DashboardPostDetailActivity` to use a shared `ApplicationScope` for file deletions rather than `GlobalScope`.
Consolidated the creation of `OkHttpClient` to reuse the existing `SharedBitmapDependencies.client`, and reused a single `Moshi` instance for both `DashboardNewsRepository` and `VoicesComposerRepository`.
Hoisted the `globalPattern` regex initialization into the companion object alongside `IMAGE_MARKDOWN_REGEX` to prevent redundant object creation.

---
*PR created automatically by Jules for task [11970576089673178704](https://jules.google.com/task/11970576089673178704) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
@@ -53,7 +53,6 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import org.ole.planet.myplanet.lite.R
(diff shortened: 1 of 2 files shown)


dogi:
**Review — 82/100** · **keeper of the post-detail scope cluster (#1079/#1110/#1120/#1126)**

**Verified:** This is the only PR touching the `clearPendingReplyImages` cleanup that gets the lifecycle right. The block is called from `onDestroy()` (line 1309) among other sites; the `lifecycleScope` variants (#1079, #1110, #1120) silently no-op there because the scope is already cancelled when the `onDestroy` body runs. `ApplicationScope.io` (SupervisorJob + IO) survives activity teardown — deletion keeps working on every path, which is exactly what the deleted `GlobalScope` was providing, now with a named, discoverable owner. The client/Moshi consolidation (three `OkHttpClient.Builder().build()` collapse into the shared client; two Moshi instances into one) and the precompiled `GLOBAL_PATTERN_REGEX` are both faithful.

**Concerns:**
- **Namespace collision with #1129:** that PR creates `org.ole.planet.myplanet.lite.ApplicationScope` (root package) for the same purpose in `CreateVoiceActivity`; this one creates `org.ole.planet.myplanet.lite.util.ApplicationScope`. If both merge, the app has two application scopes and the next contributor flips a coin. `util/` is the right home per project conventions — coordinate so #1129 imports this one.
- Mild scope-creep: the shared-deps and regex changes are separable from the scope fix; fine here since all three are correct, but it makes the eventual #1095 rebase (which touches the adjacent `actionsRepository` line) noisier.
- A one-sentence comment on `ApplicationScope.io` ("for work that must outlive the requesting component, e.g. temp-file cleanup") would stop the next person from 'fixing' call sites back to lifecycleScope.

**Verdict:** The correct resolution of the whole GlobalScope debate — merge this, close #1079/#1110/#1120, and align #1129 to reuse `util/ApplicationScope`.

---

PlataformasInformaticas:

PR #1134, opened 2026-07-03
Replace commit() with commitAllowingStateLoss() in DashboardTeamsFragment

dogi:
This PR updates `DashboardTeamsFragment` to use `commitAllowingStateLoss()` instead of `commit()` for all FragmentTransactions, reducing the risk of `IllegalStateException` after saved state.

---
*PR created automatically by Jules for task [18357201619865605145](https://jules.google.com/task/18357201619865605145) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamsFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamsFragment.kt
@@ -81,7 +81,7 @@ class DashboardTeamsFragment : Fragment(R.layout.fragment_dashboard_teams_feed) 
                 childFragmentManager.findFragmentById(R.id.teamContentContainer)?.let { fragment ->
                     childFragmentManager.beginTransaction()
                         .remove(fragment)
-                        .commit()
+                        .commitAllowingStateLoss()
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 55/100**

**Verified:** Compiles; converts the three child-fragment transactions (team-selection listener → remove/replace of voices/surveys content).

**Concerns:** Same story as the rest of the speculative quartet (#1134/#1135/#1137/#1144): no crash report, "reducing the risk" as the whole justification, and the cost is that a genuinely mistimed transaction now gets **silently dropped** — a team tap that appears to do nothing, unreproducible, with no stack trace. These transactions fire from a selection listener; if that listener can genuinely be invoked after `onSaveInstanceState` (e.g., from an async preferences callback), that's worth establishing with a stack — and then an `isStateSaved` guard (skip + log, or defer) documents the intent far better than `commitAllowingStateLoss`.

The mitigating factor, as in #1137: this child UI is rebuilt from the stored team selection on every re-entry, so dropped transactions here are recoverable. That makes the change defensible — but then say *that* in the code, once, rather than scattering silent-loss commits.

**Verdict:** Evidence-free suppression with recoverable blast radius — attach the crash or use an `isStateSaved` guard; one round. Decide once for the whole quartet.

---

PR #1135, opened 2026-07-03
Refactor: Replace commit() with commitAllowingStateLoss() in DashboardSurveysFragment

dogi:
Replaced `FragmentTransaction.commit()` with `commitAllowingStateLoss()` in `DashboardSurveysFragment.kt` to prevent potential `IllegalStateException`s if fragments are committed after the activity's state has been saved, as requested by the user.

---
*PR created automatically by Jules for task [2767229908942625560](https://jules.google.com/task/2767229908942625560) started by @dogi*


dogi:
**Review — 55/100**

**Verified:** Compiles; converts the two child-fragment transactions in the surveys container (selection-listener remove + replace).

**Concerns:** Identical critique to its siblings #1134/#1137/#1144 (the speculative `commitAllowingStateLoss` quartet): no observed crash, and the trade is a possible loud `IllegalStateException` for a guaranteed-silent dropped transaction (a surveys tab that just stays empty). Mitigation, as in #1134: this container is rebuilt from stored selection on re-entry, so loss is recoverable — which makes the change tolerable but still undocumented behavior-blurring. Prefer an `isStateSaved` guard with a comment, or attach the crash report that motivated this.

**Verdict:** Same round as the rest of the quartet — evidence or an explicit guard; decide the policy once, not per-PR.

---

PR #1137, opened 2026-07-03
Fix: Replace commit() with commitAllowingStateLoss() in DashboardActivity

dogi:
Replaced instances of FragmentTransaction.commit() with commitAllowingStateLoss() in DashboardActivity to address state loss exceptions during Fragment replacement. Unit tests pass locally.

---
*PR created automatically by Jules for task [5513625793981944983](https://jules.google.com/task/5513625793981944983) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardActivity.kt
@@ -454,7 +454,7 @@ class DashboardActivity : BaseActivity() {
         if (fragment !is DashboardSurveysFragment) {
             supportFragmentManager.beginTransaction()
                 .replace(R.id.dashboardSurveysContainer, DashboardSurveysFragment())
-                .commit()
+                .commitAllowingStateLoss()
         }
 
         updateBottomNavigationState()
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 55/100**

**Verified:** Compiles; the four converted transactions are the bottom-nav section swaps (`showSurveys`/`showTeamMembers`/`showCourses`/`showResources`).

**Concerns — this is speculative suppression, not a fix:**
- These transactions run from **bottom-nav click handlers**. Click events don't fire while the activity is stopped, and on modern androidx `onSaveInstanceState` runs at stop — so the window for the `IllegalStateException` here is between vanishingly small and nonexistent. The description says "to address state loss exceptions" but cites no crash; if there's a Crashlytics stack pointing at these lines, that belongs in the PR body and changes the calculus.
- `commitAllowingStateLoss()` isn't free: when it does trigger, the transaction is **silently dropped** — the user lands on a nav tab whose container never got its fragment, with no exception to find. In this specific activity the damage is limited (each `showX` re-replaces the container on next selection), which is why this is a 55 and not a 30 — but the honest pattern for "commit if safe" is a `supportFragmentManager.isStateSaved` guard, which preserves loud failure everywhere else.

**Cross-PR:** one of the speculative quartet (#1134/#1135/#1137/#1144 — see #1144 where the change is provably a no-op). If the team decides these are wanted as belt-and-suspenders, fine — but decide once, with a comment explaining why state loss is acceptable, not four times silently.

**Verdict:** Defensible-but-evidence-free crash suppression — provide the crash report or switch to an `isStateSaved` guard; one round.

---

PR #1138, opened 2026-07-03
Refactor: replace SharedPreferences commit() with apply()

dogi:
Replaced synchronous `SharedPreferences.Editor.commit()` calls with asynchronous `apply()` calls in `MyPlanetLite.kt` to prevent potential main thread blocking and ANR issues. Used `sed` as requested. Verified changes with build, lint, and unit tests.

---
*PR created automatically by Jules for task [16315880394183710805](https://jules.google.com/task/16315880394183710805) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/MyPlanetLite.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/MyPlanetLite.kt
@@ -1097,7 +1097,7 @@ class MyPlanetLite : BaseActivity() {
             .putBoolean(KEY_REMEMBER_CREDENTIALS, true)
             .putString(KEY_REMEMBERED_USERNAME, username)
             .putString(org.ole.planet.myplanet.lite.profile.ProfileCredentialsStore.getPasswordKey(username), password)
-            .commit()
+            .apply()
     }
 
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 78/100**

**Verified:** Both converted sites (`saveRememberedCredentials`, `clearRememberedCredentials`) write to `securePreferences` and ignore `commit()`'s return value, so the swap compiles and changes nothing observable in-process: `apply()` updates the in-memory map synchronously, and this app is single-process, so every subsequent read (including the login flow reading credentials right after saving) sees the new values immediately. What moves off the main thread is the disk persistence — which for `EncryptedSharedPreferences` rides on top of the encryption work and is exactly the kind of I/O StrictMode flags. Legitimate, tiny win.

**Trade-off worth stating (not blocking):** `apply()` gives up the durability guarantee — if the process dies before the async write lands, a just-saved "remember me" credential silently doesn't persist and the user logs in again next launch. For remember-me convenience data that's an acceptable loss; for anything session-critical it wouldn't be. Worth one sentence in the description so the next reader knows it was a decision, not an accident.

**Verdict:** Correct, appropriately scoped, honest premise — mergeable as-is.

---

PlataformasInformaticas:

PR #1133, opened 2026-07-03
Refactor FullscreenPdfActivity to use ListAdapter

dogi:
Refactored `FullscreenPdfActivity` to use `ListAdapter` instead of the base `RecyclerView.Adapter`. This improves performance by leveraging `DiffUtil` for calculating view changes when the item list updates.

---
*PR created automatically by Jules for task [12423972392231870065](https://jules.google.com/task/12423972392231870065) started by @dogi*

PR #1093, opened 2026-07-03
♻️ Refactor OutboxEntryDiffCallback to use DiffUtils

dogi:
Refactored the DiffCallback implementation in `DashboardTeamSurveysFragment` to reduce boilerplate by leveraging the `DiffUtils.itemCallback` utility method for `SurveyOutboxAdapter`.

---
*PR created automatically by Jules for task [15367712454017795678](https://jules.google.com/task/15367712454017795678) started by @dogi*

PR #1064, opened 2026-07-02
Fix Markdown Rendering for Survey Content

PlataformasInformaticas:
## Fix Markdown Rendering for Survey Content

🎯 **What:**
Integrated the Markwon library to properly parse and render Markdown in survey titles, descriptions, and question bodies throughout myPlanet Lite.

💡 **Why:**
Survey content may contain rich Markdown formatting, including headings, lists, links, emphasis, images, and multi-line text. Previously, this content was displayed as plain text, reducing readability and causing formatting issues.

🛠️ **Solution:**
- Integrated Markwon for Markdown parsing and rendering.
- Enabled Markdown rendering for survey titles, descriptions, and question content.
- Explicitly converted newline characters to Markdown line breaks (`  \n`) to preserve multi-line formatting.
- Made survey titles and descriptions in the dashboard list clickable, improving navigation and usability.
- Preserved compatibility with plain text survey content.

✅ **Verification:**
- Verified rendering of headings, bold, italic, lists, links, images, and multi-line content.
- Confirmed Markdown renders correctly across survey titles, descriptions, and questions.
- Verified dashboard titles and descriptions are clickable and behave as expected.
- Confirmed existing plain text surveys continue to display correctly without regressions.

✨ **Result:**
Survey content is now rendered with full Markdown support, providing a richer and more readable user experience while improving interaction with survey items.

**Fixes #1063**

image
image

PR #1141, opened 2026-07-03
Migrate RecyclerView.Adapter to ListAdapter for SurveysPagerAdapter

dogi:
Migrates the `SurveysPagerAdapter` in `DashboardTeamSurveysFragment.kt` to extend `androidx.recyclerview.widget.ListAdapter`. Implements `DiffUtil.ItemCallback`, removes `getItemCount()`, replaces manual array access with `getItem(position)`, and changes `enum class Page` visibility.

---
*PR created automatically by Jules for task [10775249033505154914](https://jules.google.com/task/10775249033505154914) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamSurveysFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamSurveysFragment.kt
@@ -376,9 +376,11 @@ private class SurveysPagerAdapter(
     private val onSurveyDownloadRequested: (SurveyDocument) -> Unit,
     private val onOutboxSelected: (OutboxEntry) -> Unit,
     private val onStatusChanged: () -> Unit,
-) : RecyclerView.Adapter<SurveysPagerAdapter.PageViewHolder>() {
+) : androidx.recyclerview.widget.ListAdapter<SurveysPagerAdapter.Page, SurveysPagerAdapter.PageViewHolder>(PageDiffCallback()) {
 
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 50/100**

**Verified:** Compiles, behavior identical — the pager still shows exactly `[TEAM, ADOPTED, OUTBOX]`.

**Assessment — ListAdapter over a list that cannot change.** The page list is a hardcoded constant submitted once in `init`; the new `PageDiffCallback` can never observe a difference, `submitList` is never called again, and page *content* updates still flow through the existing manual `notifyItemChanged` calls exactly as before. So the migration adds a diff callback, an init block, and a widened enum visibility (`private` → public) in exchange for nothing. Same critique as #1133 (static PDF pages): `ListAdapter` earns its keep on mutating datasets — the sibling migrations over resources/courses qualify; a fixed three-tab pager doesn't.

Also: stray double blank line after `PageDiffCallback`; the enum's visibility widening is only needed because the supertype generic mentions it — a nested sealed/enum kept `private` with a `ListAdapter<...>` supertype can't work, which is itself a hint this shape doesn't fit.

**Cluster:** same fragment as #1064 (priority) and the diff-callback cluster (#1093/#1097/#1103/#1114) — merging this creates rebase noise in the most contested file of the set, for zero functional gain.

**Verdict:** Pattern applied where it has no purchase — recommend closing; if the team wants uniformity-for-uniformity's-sake, say so explicitly in the description.

---

PR #1142, opened 2026-07-03
Refactor VoicesComposerRepository upload network sequence

dogi:
Refactored network sequence into a repository class to clean up the activity logic.

---
*PR created automatically by Jules for task [8509139866013412139](https://jules.google.com/task/8509139866013412139) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
@@ -1127,19 +1127,23 @@ class DashboardPostDetailActivity : org.ole.planet.myplanet.lite.BaseActivity() 
             return markdown
         }
 
-        val uploadResponse = composerRepository.uploadReplyImageResource(
+        val uploadResult = composerRepository.uploadReplyImageResource(
             baseUrl,
             credentials,
             context,
             pending.fileName,
-            pending.jpegBytes
-        )
-        val relativeMarkdown = uploadResponse.markdown!!
+            pending.jpegBytes,
+            pending.resourceId,
+            pending.resourceRevision
+        ) { createdId, createdRev ->
+            pending.resourceId = createdId
+            pending.resourceRevision = createdRev
+        }
 
-        pending.resourceId = uploadResponse.resourceId
-        pending.resourceRevision = uploadResponse.revision
-        pending.uploadedMarkdown = relativeMarkdown
-        return relativeMarkdown
+        pending.resourceId = uploadResult.resourceId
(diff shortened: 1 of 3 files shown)


dogi:
**Review — 62/100**

**Verified:** The create-document → upload-binary → resolve-fallbacks sequence moves into `VoicesComposerRepository.uploadReplyImageResource` mostly faithfully — resolved id/filename/revision fallbacks and the synthesized markdown fallback are identical.

**Concern — a retry-behavior regression hides in the move:**
- Old code set `pending.resourceId = creationResponse.id` (and revision) **immediately after document creation, before the binary upload**. That wasn't sloppiness — it meant a failed upload left the pending image marked with its created document, and per the guard at the top of `ensureReplyImageUpload`, association state survived for retry handling. New code sets `pending.resourceId` only **after the whole sequence succeeds** (`DashboardPostDetailActivity.kt:1138-1140`); if `uploadResourceBinary` throws, the pending image has no `resourceId`, and the next send attempt calls `createResourceDocument` again — **orphaning a duplicate resource document on the server per retry**. Either update `pending` from inside the flow (pass a callback / return the creation response on failure) or have the repository method itself retry the binary against the already-created doc.
- `uploadResponse.markdown!!` — the repository does guarantee non-null via the fallback, but encoding that guarantee as `!!` at the call site is fragile; return a small non-nullable result type from `uploadReplyImageResource` instead of reusing the nullable `ResourceUploadResponse`.
- No test, though the new method is MockWebServer-testable and the fallback resolution logic is exactly the kind of thing that silently breaks.

**Cluster:** `DashboardPostDetailActivity` again (#1079/#1108/#1110/#1120/#1121/#1126) — rebase after #1126.

**Verdict:** Good extraction shape, but it quietly changes failure/retry semantics in a way that can litter the server with orphaned docs — needs a round.

---

dogi:
**Re-review after latest push — fixed, elegantly. 78/100, lifting changes requested.**

The retry-duplication regression is resolved better than I asked: `uploadReplyImageResource` now takes the pending image's `existingResourceId`/`existingRevision` and an `onDocumentCreated` callback that updates `pending` **immediately after document creation** — so a failed binary upload leaves the created doc recorded, and the retry *reuses* it (skipping re-creation entirely) instead of orphaning duplicates. That's strictly better than the original code, which recorded the id but still had no reuse path. The `!!` is gone too — `ReplyImageUploadResult` carries a non-nullable `markdown`.

Still open (non-blocking): no test on the new method; the create-reuse-retry behavior is exactly what a MockWebServer test would lock in — worth a follow-up.

**Verdict:** Regression fixed with a genuinely improved retry story — ready.

---

PlataformasInformaticas:

PR #1143, opened 2026-07-03
Fix OOM spike during media transform output reading

dogi:
Replaces `outputFile.readBytes()` with `readBytesIfUnderSize(outputFile)` to prevent OutOfMemory errors during unexpectedly large audio or video transcoding outputs. Unit tests added for size limit guard and failure cleanups.

---
*PR created automatically by Jules for task [15281685635293499739](https://jules.google.com/task/15281685635293499739) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardResourcesMediaUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardResourcesMediaUtils.kt
@@ -48,6 +48,8 @@ import org.json.JSONArray
 import org.json.JSONObject
 
 object DashboardResourcesMediaUtils {
+    const val MAX_IN_MEMORY_SIZE_BYTES = 50L * 1024 * 1024
+
     data class ResourceUploadMetadata(
         val fileName: String,
         val defaultTitle: String,
@@ -400,7 +402,7 @@ object DashboardResourcesMediaUtils {
                 continuation.invokeOnCancellation { transformer.cancel() }
                 transformer.start(editedMediaItem, outputFile.absolutePath)
             }
-            if (!exportSucceeded) null else withContext(Dispatchers.IO) { outputFile.readBytes() }
+            if (!exportSucceeded) null else withContext(Dispatchers.IO) { readBytesIfUnderSize(outputFile) }
         } catch (_: Throwable) {
             null
         } finally {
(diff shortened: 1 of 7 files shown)


dogi:
**Review — 65/100**

**Verified:** The mechanics are sound — `readBytesIfUnderSize` throws inside the transcode functions' `catch (_: Throwable) → null` envelope, so an oversized output now degrades to the normal failure path instead of `OutOfMemoryError`. The cap tests are real, and the two cleanup tests legitimately pin that failed transcodes leave no temp files behind.

**Concerns:**
- **The cap trades OOM for a functional regression.** A user who transcodes a long video into, say, a 60MB MP4 *used to succeed* (a 60MB `ByteArray` is usually fine); now they get a generic failure toast with no explanation, and nothing in the UI says "too large." Two issues layered here: (1) 50MB is arbitrary and undocumented — why not 32 or 100? — and (2) the honest fix for "don't hold the whole output in memory" is **streaming the upload from the file** (OkHttp `RequestBody` from a `File` — the upload path already has the file on disk!) rather than refusing to read it. If streaming is out of scope, at least surface a specific "file too large" message instead of routing through `catch(Throwable) → null`.
- `readBytesIfUnderSize` throwing `IllegalStateException` as control flow, only to be swallowed by a blanket catch two lines up, is a roundabout way to write `if (file.length() > max) return null`.
- The two cleanup tests pass by triggering the failure path via mocked `Context` — fine, but they're unrelated to the OOM cap this PR is named for; the cap is never exercised through a real transcode.

**Verdict:** Reasonable stopgap with real tests, but it silently converts large-but-valid outputs into unexplained failures — either stream the upload or surface the limit to the user; one round.

---

dogi:
**Re-review after latest push — fixed with the real solution. 78/100, lifting changes requested.**

Instead of patching the 50MB cap, the push implements what I asked for: transcodes now return the output **`File`**, a sealed `UploadMedia` (Bytes | FileMedia) flows through `performResourceCreateAndUpload`, and the repository streams file-backed uploads — so transformed media never has to fit in a `ByteArray` at all. The arbitrary cap, the exception-as-control-flow helper, and its tests are all gone; the temp-file lifecycle is handled (output kept on success for the upload, deleted on failure, and the upload path wrapped in try for cleanup). Large-but-valid videos upload again instead of silently failing — the functional regression I flagged is not just avoided but structurally impossible now.

Remaining nit: image/PDF paths still go through `UploadMedia.Bytes` (fine — they're bounded by the resize step), and the cleanup-on-failure tests survived the refactor.

**Verdict:** The streaming fix, done properly — ready.

---

PlataformasInformaticas:

PR #1144, opened 2026-07-03
Use commitAllowingStateLoss() in SurveyWizardActivity

dogi:
Replaced `commit()` with `commitAllowingStateLoss()` for the fragment transaction in `SurveyWizardActivity.kt` to prevent `IllegalStateException`s if the fragment is committed after the activity state has been saved.

---
*PR created automatically by Jules for task [16529222676855689691](https://jules.google.com/task/16529222676855689691) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/SurveyWizardActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/SurveyWizardActivity.kt
@@ -54,7 +54,7 @@ class SurveyWizardActivity : AppCompatActivity() {
                         intent.getBooleanExtra(EXTRA_INCLUDE_USER_CONTEXT, true),
                     ),
                 )
-                .commit()
+                .commitAllowingStateLoss()
         }
     }
 


dogi:
**Review — 40/100**

**Verified — the exception this prevents cannot occur here.** The only transaction in `SurveyWizardActivity` sits in `onCreate()` inside `if (savedInstanceState == null)` (`SurveyWizardActivity.kt:43-57`). `commit()` throws "Can not perform this action after onSaveInstanceState" only when the state has been saved — during `onCreate` it never has been. This one-character change is a no-op safety blanket for an impossible crash, and its cost is real: `commitAllowingStateLoss()` tells every future reader "this transaction may be silently dropped," which is false and misleading for the activity's *initial content* — the one transaction you'd absolutely want to fail loudly if it somehow didn't run.

**Recommendation:** close this PR (or revert to `commit()`). If there's an actual crash report with this stack, attach it and the discussion changes.

**Cross-PR:** part of the speculative quartet #1134/#1135/#1137/#1144, all swapping `commit()`→`commitAllowingStateLoss()` "to reduce risk" with no observed crash. Same critique applies throughout; this instance is the clearest-cut no-op of the four.

**Verdict:** Fixes an impossible bug while blurring the code's intent — close.

---

PR #1145, opened 2026-07-03
Refactor CourseAdapter to use AsyncListDiffer for better performance

dogi:
**What:** Refactored `CourseAdapter` to use `AsyncListDiffer` for optimized list updates.
**Why:** To ensure smooth UI updates and avoid costly manual notifyDatasetChanged/range notifications for dynamic filtering or appends, while keeping the external API unaffected.
**How:** 
1. Replaced the manually managed `displayedItems` with an `AsyncListDiffer` backed by a customized `DiffUtils.itemCallback`. 
2. Supplied a custom `ListUpdateCallback` to shift payload positional updates by +1, preserving the static header view's stability.
3. Updated external functions (submitCourses, appendCourses, updateCourseProgress, updateDownloadedCourses, applyFilter) to invoke list state changes against the `differ` instead.

---
*PR created automatically by Jules for task [10652789879932138311](https://jules.google.com/task/10652789879932138311) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
@@ -1171,11 +1171,34 @@ class DashboardCoursePageFragment : Fragment(R.layout.fragment_dashboard_courses
         private val imageLoaderProvider: () -> DashboardPostImageLoader?,
         private val ensureImageLoader: () -> Unit,
         private val categoriesProvider: () -> List<CourseCategory>
-    ) :
-        RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+    ) : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+
+        private val differ = androidx.recyclerview.widget.AsyncListDiffer(
+            object : androidx.recyclerview.widget.ListUpdateCallback {
+                override fun onInserted(position: Int, count: Int) {
+                    notifyItemRangeInserted(position + 1, count)
+                }
+
+                override fun onRemoved(position: Int, count: Int) {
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 70/100**

**Verified:** The most conservative of the four rewrites — keeps `RecyclerView.Adapter` and the virtual position-0 header, and swaps the hand-rolled `DiffUtil.calculateDiff` for an `AsyncListDiffer` whose `ListUpdateCallback` re-applies the +1 header offset. The offset callbacks are implemented correctly (insert/remove/move/change all shifted), diffing moves off the main thread, and behavior is otherwise preserved — including keeping the header always present (`currentList.size + 1`), which the other three rewrites accidentally lose. `updateDownloadedCourses`/`updateDownloadProgress` keep manual notifies, which is *consistent* here since the adapter never claims full ListAdapter semantics.

**Concerns:**
- Internally consistent but still two update mechanisms (differ for data, manual notifies for download state); a rapid `submitList` diff landing between a manual `notifyItemChanged` and layout can still misalign positions — the +1-offset flavor of the same race the old code had.
- The diff sprays double blank lines through the adapter (lines ~1213, 1219, 1241, 1265, 1287...) — sloppy.
- Title says "for better performance"; the old code already used `DiffUtil.calculateDiff` — the gain is *async* diffing only, worth claiming precisely.

**Cross-PR:** one of four competing rewrites (#1098/#1127/#1140/#1145). #1098's state-in-item design eliminates the dual-mechanism problem entirely and is the recommended keeper; this is the best fallback if the team wants the minimal-risk option. **Don't merge more than one.**

**Verdict:** Correct, minimal, a bit scruffy — close in favor of #1098 unless the team prefers minimal risk over full migration.

---

PR #1147, opened 2026-07-03
feat: debounce invite member remote search by 300ms

dogi:
debounce invite member remote search by 300ms

Added a debounce to the search input in the invite members dialog.
This prevents fetching the page aggressively while the user is typing
and waiting 300ms before triggering the load, saving network bandwidth
and improving performance without changing paging behavior.

---
*PR created automatically by Jules for task [8661618514762178028](https://jules.google.com/task/8661618514762178028) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamMembersSupport.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardTeamMembersSupport.kt
@@ -22,6 +22,8 @@ import androidx.recyclerview.widget.RecyclerView
 import com.google.android.material.floatingactionbutton.FloatingActionButton
 import kotlin.math.abs
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
 import org.ole.planet.myplanet.lite.dashboard.DashboardAvatarLoader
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 66/100**

**Verified:** Correct debounce — and notably, this version gets the ordering right that its twin #1122 gets wrong: `currentSearchTerm = newQuery` stays synchronous in the listener, so the `newQuery != currentSearchTerm` guard never compares against a stale value (in #1122 a quick type-and-revert leaves an orphaned search job whose results land under an empty box; here it's correctly cancelled).

**Concerns:**
- **Duplicate of #1122** (same listener, same 300ms, same `searchJob` pattern). That PR additionally cancels the pending job on dialog dismiss (this one leaks a scheduled `loadInvitePage` past dismissal — mostly harmless since `onReload` refreshes anyway, but sloppier) and ships a virtual-time test proving the debounce. **Recommended resolution: keep #1122, port this PR's synchronous `currentSearchTerm` assignment into it (one line), close this one.**
- No test here; #1122's test would have caught nothing for you, but writing none means neither the debounce nor the cancel is pinned.

**Verdict:** Right core ordering, but the twin with the test and the dismiss-cancel is the better base — close in favor of a fixed #1122.

---

PR #1150, opened 2026-07-03
Refactor DashboardCoursePageFragment to reduce header recreation

dogi:
Refactor DashboardCoursePageFragment to reduce header recreation

Decoupled the header state from row list updates by introducing internal state caching within HeaderViewHolder. Prevented full recreation of the category chip group by only clearing and rebuilding it when the list of categories actually changes. Reused the search listener lifecycle to minimize listener churn on every rebind. Finally, removed the redundant notifyItemChanged(0) call in updateCategories to stop forcing a full header rebind.

---
*PR created automatically by Jules for task [6657724608036661455](https://jules.google.com/task/6657724608036661455) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
@@ -1221,6 +1221,7 @@ class DashboardCoursePageFragment : Fragment(R.layout.fragment_dashboard_courses
             if (selectedCategory >= categoriesProvider().size) {
                 selectedCategory = 0
             }
+            notifyItemChanged(0)
         }
 
         fun updateTagFilter(courseIds: Set<String>?) {
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardCoursePageFragment.kt
@@ -1207,7 +1207,6 @@ class DashboardCoursePageFragment : Fragment(R.layout.fragment_dashboard_courses
                 holder.onCategorySelected = onCategorySelected
                 holder.bind(selectedCategory, categoriesProvider(), searchQuery, { index ->
                     selectedCategory = index
-                    notifyItemChanged(0)
                 }) { query ->
                     updateSearchQuery(query)
                 }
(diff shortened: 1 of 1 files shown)


dogi:
**Review — 42/100**

**Verified — the optimization breaks category updates.** The in-holder diffing (`currentCategories != categories` → only rebuild chips on change; listeners created once in `init`) is genuinely good work. But the PR also deletes **both** `notifyItemChanged(0)` calls, including the one in `updateCategories()` (`DashboardCoursePageFragment.kt:1221`) — and that notify was the *only* trigger that ever re-bound the header after the categories load. Sequence on a fresh screen: RecyclerView binds the header once while `categoriesProvider()` is still empty → categories arrive → fragment calls `adapter.updateCategories()` → **nothing notifies** → `bind` never runs again → the chip group stays empty until some unrelated event happens to rebind position 0 (e.g. the header scrolling out of the viewport and back). The whole point of making `bind` cheap-when-unchanged is that you can keep the notify and let the holder no-op it — removing the notify instead removes the update path.

**Fix:** restore `notifyItemChanged(0)` in `updateCategories()` (the new caching makes it nearly free). Removing the one inside the chip-selection callback is fine — the holder now maintains its own checked state.

**Also:**
- `currentSelectedIndex` can go stale when `updateCategories()` clamps `selectedCategory` to 0 with no rebind — same root cause.
- **Cluster:** this rewrites `HeaderViewHolder` inside the same adapter that #1098/#1127/#1140/#1145 rewrite wholesale and #1124/#1132 patch — this PR conflicts with all of them. If #1098 lands as recommended, this needs a full rebase.

**Verdict:** Good caching idea sabotaged by deleting the update trigger it was meant to make cheap — needs the notify restored; one round.

---

dogi:
**Re-review after latest push — fixed. 74/100, lifting changes requested.**

`notifyItemChanged(0)` is restored in `updateCategories()` — with the in-holder caching the rebind is nearly free and the update path works again (the deleted notify inside the chip callback correctly stays deleted). Standing caveat: this PR and #1098 are competing solutions to the same header-churn problem (holder-level caching vs. state-in-item) on the same adapter — if #1098 merges as cluster keeper, close this one rather than rebase.

**Verdict:** fixed — ready.

---

PlataformasInformaticas:

PR #1151, opened 2026-07-03
feat: Constrain shared image loader coroutine scopes with cleanup APIs

dogi:
Constrain shared image loader coroutine scopes. Added explicit production-safe clear/cancel APIs to `DashboardAvatarLoader` and `DashboardPostImageLoader` for in-flight image requests. Ensured that owners call these cleanups during teardown, maintaining symmetry between both loaders to avoid conflicts. Made companion objects public to allow correct API usage.

---
*PR created automatically by Jules for task [7466467186413102095](https://jules.google.com/task/7466467186413102095) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/DashboardVoicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/DashboardVoicesFragment.kt
@@ -256,7 +256,10 @@ class DashboardVoicesFragment : Fragment(R.layout.fragment_dashboard_voices) {
         fetchJob = null
         recyclerView.adapter = null
         avatarLoader?.destroy()
+        DashboardAvatarLoader.clearInFlightRequests()
         avatarLoader = null
+        postImageLoader?.destroy()
+        DashboardPostImageLoader.clearInFlightRequests()
         postImageLoader = null
         postShareHelper = null
         AvatarUpdateNotifier.unregister(avatarUpdateListener)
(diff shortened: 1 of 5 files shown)


dogi:
**Review — 25/100**

**Verified — this introduces a deterministic cross-screen cancellation bug.** `inFlightRequests` and `inFlightScope` are **static**, shared by every loader instance (that's the point of the current design — cross-screen dedupe). The new `clearInFlightRequests()` is also static, but it's called from *individual screens'* teardown (`DashboardVoicesFragment.onDestroyView`, `DashboardPostDetailActivity.onDestroy`, `DashboardTeamMemberProfileActivity.onDestroy`). Net effect: closing **any one** of these screens cancels **every** in-flight avatar/post-image fetch app-wide, including those belonging to the screen the user is returning to. Open a member profile from the members list, close it → the members list behind it silently stops loading avatars. The description's "maintaining symmetry between both loaders" makes both loaders wrong symmetrically.

**Also:**
- `DashboardPostImageLoader.destroy() {}` (`DashboardPostImageLoader.kt:80`) — an empty method added so call sites have something to call. Delete it or give it a body that means something.
- **Mutually exclusive with #1082/#1083 (as revised):** those PRs move `inFlightRequests` to instance level, which removes the static map, the static scope, and any need for this API. That design solves the ownership problem correctly; this one globalizes it.

**Recommendation:** close in favor of #1082 + #1083.

**Verdict:** Global kill-switch wired to local lifecycles — do not merge.

---

PR #1157, opened 2026-07-21
Optimize CourseAdapter image loader invalidation

dogi:
Optimized the invalidation of the `CourseAdapter` when the image loader is ready. Instead of calling `notifyDataSetChanged()`, which refreshes the entire list, we now use a targeted helper function `notifyImageLoaderReady()` that only notifies items containing image URLs, preventing unnecessary rebinds of headers or text-only rows.

---
*PR created automatically by Jules for task [9625782590446745761](https://jules.google.com/task/9625782590446745761) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course image updates when the image loader becomes ready.
  * Course listings now refresh only the relevant items, helping images appear more reliably and reducing unnecessary UI refreshes.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/lite/CourseAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/CourseAdapter.kt
@@ -156,6 +156,14 @@ class CourseAdapter(
         applyFilter()
     }
 
+    fun notifyImageLoaderReady() {
+        currentList.forEachIndexed { index, item ->
+            if (item is CourseListItem.CourseItemWrapper && !item.course.coverPath.isNullOrBlank()) {
+                notifyItemChanged(index)
+            }
+        }
+    }
+
     fun isHeader(position: Int) = getItem(position) is CourseListItem.Header
 
     private fun updateSearchQuery(query: String) {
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 90/100**

Verified against source. `CourseAdapter` is a `ListAdapter`, so `currentList` indices map directly to adapter positions; iterating and calling `notifyItemChanged(index)` only for `CourseItemWrapper` rows with a non-blank `coverPath` is exactly right — `CourseViewHolder.bind()` (CourseAdapter.kt:~327) invokes `loader.bind(imageView, coverPath)` only for those rows, so headers/text-only rows genuinely have nothing to rebind. Replacing the blanket `notifyDataSetChanged()` avoids rebinding rows that can't change while still refreshing every image row once the loader is set. The description's claim is accurate. Build + test are green.

**Strengths**
- Correct index space (ListAdapter position == currentList index), no off-by-one risk.
- The guard predicate matches the exact condition under which `bind()` uses the loader, so no image row is missed and none is needlessly touched.
- Enables item-level RecyclerView behavior instead of a full invalidation.

**Concerns**
- None blocking. Minor: this is a one-shot refresh after the loader becomes ready, so item-change animations are irrelevant here; no need for a payload. Fine as-is.

**Verdict:** Correct, minimal, measurably better than `notifyDataSetChanged()` — mergeable. (Author == reviewer here, so a second reviewer is still needed to approve.)

---

dogi:
**Update after reviewing #1170 — relabeling `duplicate` (was `ready`).**

This PR and #1170 change the **same line** (`DashboardCoursePageFragment.kt:255`, the `adapter.notifyDataSetChanged()` in the image-loader-ready block) for the same purpose. They can't both land, so this is a duplicate of #1170.

#1170 is the stronger implementation: it introduces an `IMAGE_UPDATE_PAYLOAD` and re-binds via `notifyItemRangeChanged(..., IMAGE_UPDATE_PAYLOAD)` so only `bindImage` runs — a payload-scoped refresh that avoids the full per-row rebind this PR's payload-less `notifyItemChanged(index)` performs, and it explicitly fixes image flicker. It also composes with #1169's `HEADER_UPDATE_PAYLOAD`.

Nothing is wrong with the code here — my earlier 90/ready assessment stands on its own merits — but to avoid two conflicting fixes I recommend **closing this in favor of #1170**. Marking `duplicate` and lifting `ready`.

---

PR #1156, opened 2026-07-21
Limit concurrency for comment image fetches in DashboardPostDetailActivity

dogi:
This bounds the concurrent async fetches to match the pattern used in
CreateVoiceActivity, preventing potential OOM issues or thread pool
exhaustion while avoiding new global architectures.

---
*PR created automatically by Jules for task [13241983937574728541](https://jules.google.com/task/13241983937574728541) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading comment images by limiting how many image fetches run at the same time.
  * Reduces the risk of slowdowns and resource spikes during reply image loading.
* **Chores / Release**
  * Updated the app version (version code/name) for this release.

PR #1162, opened 2026-07-21
Refactor reply image upload into VoicesComposerRepository

dogi:
Refactored the network sequence for replying to a post with an image out of the UI activity (`DashboardPostDetailActivity`) and into `VoicesComposerRepository` (`ensureReplyImageUpload`). This correctly aligns with our repository-pattern refactoring goals. Tests passing and lint applied.

---
*PR created automatically by Jules for task [6254076356421938842](https://jules.google.com/task/6254076356421938842) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reply voice image attachments by avoiding duplicate uploads when the image markdown is already available.
  * Ensured reply images reliably retain/resolve resource identifiers and revisions, with correct generated image markup.

* **Tests**
  * Added unit test coverage for the two-step reply image flow: resource creation followed by binary upload, including correct request headers and generated markdown output.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -34,8 +34,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 487
-        versionName = "0.4.87"
+        versionCode = 519
+        versionName = "0.5.19"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailReplyActionExtensions.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailReplyActionExtensions.kt
@@ -241,28 +241,18 @@ internal suspend fun DashboardPostDetailActivity.ensureReplyImageUpload(
         return markdown
     }
 
-    val metadata = VoicesComposerRepository.ResourceMetadataRequest.fromContext(context, pending.fileName)
-    val creationResponse = composerRepository.createResourceDocument(baseUrl, credentials, metadata)
-    pending.resourceId = creationResponse.id
-    pending.resourceRevision = creationResponse.revision
-    val uploadResponse =
-        composerRepository.uploadResourceBinary(
+    val result =
+        composerRepository.ensureReplyImageUpload(
             baseUrl,
             credentials,
-            creationResponse.id,
+            context,
             pending.fileName,
-            creationResponse.revision,
             pending.jpegBytes,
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 72/100**

The extraction itself is clean and the new test (`ensureReplyImageUpload correctly coordinates creation and upload`) is genuinely good — it asserts the POST→PUT ordering, the `If-Match: rev_creation` header, and the resolved id/rev/markdown. Build + test are green. But this is **not** the pure refactor the description implies — it changes the partial-failure retry semantics.

**Concern (behavioral) — `DashboardPostDetailReplyActionExtensions.kt:237-260`**
The old code set `pending.resourceId = creationResponse.id` / `resourceRevision` *immediately after creation, before* `uploadResourceBinary`. The new code sets `pending.resourceId` only *after* the whole repository call returns, alongside `uploadedMarkdown`. Consequences:
1. **The idempotency guard at lines 237-242 (`if (existingResourceId != null)`) is now effectively dead code.** `pending.resourceId` is only ever assigned together with `pending.uploadedMarkdown`, so the earlier `pending.uploadedMarkdown?.let { return it }` (line 236) always short-circuits first. The `resourceId`-only branch can no longer be reached in the in-memory flow.
2. **Create-succeeds-but-upload-fails retry changes.** Old: `resourceId` was left set → retry returned markdown for a resource whose binary never uploaded (a latent *bug* — broken image). New: `resourceId` stays null → retry re-creates + re-uploads (working image, but orphans the first empty resource doc).

`PendingVoiceImage` is in-memory only (holds a `File` + `ByteArray`), so this is a same-session retry path, not persistence — but it's still a real semantic shift the PR doesn't call out.

**Ask:** the new retry behavior is arguably the better one, so I'd lean *keep it* — but then delete or rework the now-unreachable `existingResourceId != null` branch so the code doesn't advertise an idempotency path it no longer has. Either way, make the choice intentional rather than incidental to the extraction.

**Cross-PR:** #1156 edits the same file (`loadExistingCommentImages`) — different function, but expect a trivial rebase for whichever lands second.

**Verdict:** Good extraction + test, but resolve the dead idempotency guard / document the retry-semantics change before merge — one more round.

---

dogi:
**Re-review (commit `b4f3569d`) — fixed → ready**

My flag is resolved: the dead `if (existingResourceId != null) { … }` idempotency guard is removed. The only short-circuit now is `pending.uploadedMarkdown?.let { return it }`, and `resourceId`/`revision`/`uploadedMarkdown` are all set together from the `ensureReplyImageUpload` result — so there's no longer an unreachable branch advertising idempotency it doesn't have. The create-then-upload-fail retry now cleanly re-creates+re-uploads, which is the intended behavior. Repository extraction + the `ensureReplyImageUpload` test are unchanged and correct. Build + test green.

Lifting `changes requested` → `ready`. (Author == reviewer; second approver needed.)

---

PlataformasInformaticas (approved):

PR #1155, opened 2026-07-21
Refactor to reuse shared dependencies in dashboard components

dogi:
- Reuse `SharedBitmapDependencies.client` and `AuthDependencies.moshi` in `DashboardVoicesFragment`.
- Delegate `PostShareHelper.client` to `SharedBitmapDependencies.client.newBuilder()` to maintain custom timeouts while reusing the connection pool.

---
*PR created automatically by Jules for task [9652291903386665528](https://jules.google.com/task/9652291903386665528) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading dashboard voices/news content and downloading shared images.
  * Standardized the shared networking and JSON handling used for these flows, while keeping existing connection/read timeout behavior unchanged.

PR #1165, opened 2026-07-21
Refactor DashboardPostDetailActivity for shared client and ApplicationScope

dogi:
Refactored `DashboardPostDetailActivity` to use a shared `OkHttpClient` instance via `SharedBitmapDependencies.client`, significantly reducing the number of unnecessary client initializations. Reused a single `Moshi` instance across repositories. Removed the delicate `GlobalScope` by migrating background temporary file cleanup tasks to a new `ApplicationScope.io`. Hoisted the frequently instantiated `globalPattern` regex up to the companion object to improve performance.

---
*PR created automatically by Jules for task [15389988588790886545](https://jules.google.com/task/15389988588790886545) started by @dogi*

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -34,8 +34,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 487
-        versionName = "0.4.87"
+        versionCode = 517
+        versionName = "0.5.17"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
--- a/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/lite/dashboard/DashboardPostDetailActivity.kt
@@ -43,18 +43,20 @@ class DashboardPostDetailActivity : org.ole.planet.myplanet.lite.BaseActivity() 
     internal lateinit var recyclerView: RecyclerView
     internal lateinit var loadingView: View
 
+    private val reusedMoshi = Moshi.Builder().addLast(KotlinJsonAdapterFactory()).build()
+
     internal val repository =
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 90/100**

Verified all four changes against source:
1. **Shared client** — `SharedBitmapDependencies.client` is a plain `OkHttpClient`, so replacing the three `OkHttpClient.Builder().build()` instances is behavior-equivalent and now shares one connection pool. The repos just receive it as a constructor arg, no interceptor/timeout assumptions lost.
2. **Reused Moshi** — `reusedMoshi` is declared *above* `repository`/`composerRepository` (line 46 before 48), so Kotlin's top-down property init order is satisfied; config (`addLast(KotlinJsonAdapterFactory())`) is identical to the two instances it replaces.
3. **`GLOBAL_PATTERN` hoist** — the companion regex string `"(!\\[[^]]*]\\()(.*?)(\\))"` is byte-identical to the removed local `globalPattern`, and moving it out of `transformReplyMarkdownForPreview` avoids recompiling the pattern on every preview render. Real, if small, win.
4. **`ApplicationScope.io` for cleanup** — `ApplicationScope.io` is `CoroutineScope(SupervisorJob() + Dispatchers.IO)`, so the fire-and-forget temp-file delete still runs on IO, but drops the `@OptIn(DelicateCoroutinesApi)`/`GlobalScope` — a legitimate improvement.

Build + test green.

**Strengths**
- Four independent, each-correct micro-improvements; property init order and regex equivalence both handled correctly (easy things to get subtly wrong).
- Removing `GlobalScope` in favor of a shared app scope is the right call for leak-safety.

**Concerns**
- None blocking. Consistency nit: #1155 uses `AuthDependencies.moshi` for the same purpose; you could reuse that shared instance instead of a per-activity `reusedMoshi`, but this is fine.
- **Cross-PR:** shares theme/files with #1155 (SharedBitmapDependencies), #1156 & #1162 (same `DashboardPostDetail*` cluster) — no line conflicts, but land them as a group and rebase-check.

**Verdict:** Correct, behavior-preserving code-health bundle — mergeable. (Author == reviewer; second approver needed.)

---

PlataformasInformaticas (approved):

PR #1166, opened 2026-07-21
Inject shared dependencies into DashboardResourcesRepository

dogi:
Replaces internal instantiations of `OkHttpClient` and `Moshi` in `DashboardResourcesRepository` with constructor parameters that default to the shared `AuthDependencies` instances. Also introduces an `ioDispatcher` constructor parameter to replace hardcoded `Dispatchers.IO` calls, improving testability. The API remains backwards compatible.

---
*PR created automatically by Jules for task [1289849921520368890](https://jules.google.com/task/1289849921520368890) started by @dogi*

## Summary by CodeRabbit

- **New Features**
  - Improved resource upload workflows, including document creation, file attachments, and optional team links.
  - Added resource file downloads with progress reporting.
  - Added PDF downloads to local cache, including handling existing local files and authenticated requests.
- **Bug Fixes**
  - Improved handling of unsuccessful download responses and missing required resource details.
- **Chores**
  - Updated app version to **0.4.92** (version code **492**).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -34,8 +34,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 487
-        versionName = "0.4.87"
+        versionCode = 493
+        versionName = "0.4.93"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -34,8 +34,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 493
-        versionName = "0.4.93"
+        versionCode = 492
+        versionName = "0.4.92"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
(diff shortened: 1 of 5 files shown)


dogi:
**Review — rating: 89/100**

Verified against source, including the non-obvious parts:
- **Moshi ordering preserved.** `AuthDependencies.moshi.newBuilder().add(DateStringAdapter()).build()` looks like it might append `DateStringAdapter` *after* the reflective `KotlinJsonAdapterFactory` (which would shadow it), but Moshi's `Builder.add()` inserts at `lastOffset` — before any `addLast`-ed factory. `newBuilder()` carries `lastOffset` over, so the result is `[DateStringAdapter, KotlinJsonAdapterFactory]`, identical to the old inline builder. Adapter resolution is unchanged.
- **Injection is complete.** The only two `withContext(Dispatchers.IO)` in the file (lines 250, 591) are both converted to `ioDispatcher`; nothing hardcoded is left behind.
- **No caller breaks.** Both production call sites (`FullscreenPdfActivity`, `DashboardResourcesPageFragment`) use `DashboardResourcesRepository()` with defaults. The test's previously-positional `DashboardResourcesRepository(UnconfinedTestDispatcher())` is correctly changed to the named `ioDispatcher = …` — necessary because `client` is now the first parameter. Good catch; a positional call would otherwise have silently bound the dispatcher to `client`.

Build + test green.

**Strengths**
- Correct handling of the Moshi `add`/`addLast` subtlety and the constructor-reorder test fix — both are easy to get wrong.
- Production behavior preserved (shared client/moshi, IO dispatcher), testability improved — matches the codebase's DI-by-constructor convention.

**Concerns**
- None blocking. Part of the shared-deps cluster (#1155/#1165) — no line conflicts.

**Verdict:** Correct, complete, convention-aligned injection — mergeable. (Author == reviewer; second approver needed.)

---

PlataformasInformaticas (approved):

PR #1167, opened 2026-07-21
Fix memory spike during large transformed media reading

dogi:
Fix memory spike when reading excessively large transformed media. Adds a maximum transformed file size boundary guard before calling read bytes and properly manages file closure to prevent temporary file leaking.

---
*PR created automatically by Jules for task [2687175861916573144](https://jules.google.com/task/2687175861916573144) started by @dogi*

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -34,8 +34,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 487
-        versionName = "0.4.87"
+        versionCode = 503
+        versionName = "0.5.3"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 82/100**

The premise is legitimate — `outputFile.readBytes()` loads the whole transformed file into a single `ByteArray`, which is a real memory spike for large media, and the old `copyUriToTempFile` leaked its temp file when `openInputStream` returned null/threw. Both are addressed correctly. Build + test green.

**Verified**
- The try/finally restructure (nullable `inputFile`/`outputFile` created *inside* the try, null-safe deletes in `finally`) is correct and fixes the pre-existing leak where an output-temp-file creation failure would strand the input temp file.
- `copyUriToTempFile` now `delete()`s and rethrows on exception — proper cleanup.
- `isTransformedFileOversized` uses strict `>`, and `testIsTransformedFileOversized` locks the boundary (maxSize → false, maxSize+1 → true). Good.

**Concerns**
- **Design — silent failure for legitimately large output (DashboardResourcesMediaUtils.kt:408 / :475).** A successful transform whose output exceeds 50 MB now returns `null`, indistinguishable from a transform *failure*. On a low-end "lite" device this cap is defensible, but the user gets a generic error for a video that actually transcoded fine. Please confirm this is intended, and consider surfacing a distinct "file too large" message rather than folding it into the null-failure path. (A streaming upload would avoid the spike without the cap, but that's a larger change.)
- **Behavior change (minor):** moving `copyUriToTempFile`/`createTempFile` inside the `try { … } catch (_: Throwable) { null }` means input-prep exceptions (e.g. the `readFileErrorMessage` `IllegalStateException`) are now swallowed to `null` instead of propagating. This is more uniform, but if any caller relied on that exception for a specific message, it's gone.
- **Test robustness (minor):** `copyUriToTempFile_cleansUpOnException` asserts on the *file count* of the shared `java.io.tmpdir`, which is sensitive to anything else writing there during the run. Asserting that the specific `resource_input_video*.tmp` no longer exists would be more deterministic.

**Verdict:** Correct, well-tested mitigation of a real memory/leak issue — mergeable, but please confirm the 50 MB silent-drop is the intended product behavior. (Author == reviewer; second approver needed.)

---

PlataformasInformaticas (approved):

PlataformasInformaticas (approved):

PR #1168, opened 2026-07-21
Delegate ProfileActivity birth-date helpers to DateUtils

dogi:
This change delegates the internal birth date logic of ProfileActivity into the shared DateUtils utility, ensuring pure-function behavior and adding explicit unit testing for those behaviors.

---
*PR created automatically by Jules for task [7406937200837409020](https://jules.google.com/task/7406937200837409020) started by @dogi*

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -34,8 +34,8 @@ android {
         applicationId = "org.ole.planet.myplanet.lite"
         minSdk = 28
         targetSdk = 36
-        versionCode = 487
-        versionName = "0.4.87"
+        versionCode = 495
+        versionName = "0.4.95"
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "PLANET_BASE_URL", "\"http://10.82.1.30/\"")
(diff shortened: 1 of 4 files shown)


dogi:
**Review — rating: 88/100**

Verified the moved logic against source. `parseBirthDateToMillis`, `extractBirthYearFromIso`, and `calculateAgeFromIso` are verbatim moves (only `Build.VERSION.SDK_INT` → the testable `sdkInt`). The delicate one is `formatBirthDate`: parameterizing `targetPattern` (default `"yyyy-MM-dd"`, so the existing 2-arg callers are untouched) and calling it with `"MMM d, yyyy"` reproduces the old `formatBirthDateImpl` exactly —
- SDK≥O: old `LocalDate.parse(raw).atStartOfDay(UTC).toInstant().atZone(UTC).toLocalDate()` round-trips to the same `LocalDate` as the new direct `LocalDate.parse(value)`; formatter pattern + `Locale.getDefault()` match.
- pre-O: identical input-pattern list and `SimpleDateFormat`(UTC) output.

The new `sdkInt`-based tests exercise both the pre-O and O branches for all three helpers. Build + test green.

**Strengths**
- Genuine consolidation into the well-tested `util/` layer (per CLAUDE.md's guidance to prefer extending there), with real branch coverage added.
- The `targetPattern` default keeps existing `formatBirthDate` callers behavior-identical — good backward-compat instinct.

**Concerns**
- None blocking. Two nits:
  1. `formatBirthDateImpl` now returns the raw whitespace for a whitespace-only `raw` (old code returned `""`), because `fallback = raw ?: ""` and `formatBirthDate` returns the fallback on blank input. Irrelevant for real ISO dates, but if you want byte-identical behavior, pass `""` as the fallback.
  2. `DateUtils.sdkInt` is a shared mutable var with no `@After` reset; the sdk-dependent tests each set it explicitly so they're fine, but a `@Before`/`@After` reset would prevent future ordering surprises.

**Verdict:** Correct, behavior-preserving extraction with good coverage — mergeable. (Author == reviewer; second approver needed.)

---

PlataformasInformaticas (approved):

PR #1159, opened 2026-07-21
Extract join request enrichment logic to repository layer

dogi:
Extracted user profile lookup and merging for team join requests out of the dashboard fragment and into the repository layer. The fragment now receives a fully enriched model out-of-the-box. Tests have been written to cover the network repository logic mapping and fallback handling.

---
*PR created automatically by Jules for task [4461920383103890043](https://jules.google.com/task/4461920383103890043) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Improved the team join-request list with clearer member names, usernames, and avatar availability.
  * Join requests are now sorted alphabetically by member name.
  * Added graceful handling when member profile information is unavailable.
  * Empty join-request states are displayed consistently, including when no team planet is configured.

* **Tests**
  * Added coverage for successful join-request details loading and profile-fetch failures.

PR #1171, opened 2026-07-21
🧪 [testing improvement] Add test for DashboardTeamsOperations network failure

dogi:
🎯 **What:** Added a missing try/catch test for network request failures in `DashboardTeamsOperations`.
📊 **Coverage:** The private method `fetchUserProfile` is now tested for network failure by using `MockWebServer` with `SocketPolicy.DISCONNECT_AFTER_REQUEST`. It asserts that when the network request throws an `IOException`, the method catches it and returns `null`.
✨ **Result:** Increased test coverage for error conditions.

---
*PR created automatically by Jules for task [17675947195213747276](https://jules.google.com/task/17675947195213747276) started by @dogi*

## Summary by CodeRabbit

* **Tests**
  * Added coverage for handling network failures while fetching a user profile.
  * Verified that a disconnected network request produces a null result instead of causing unexpected behavior.

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/lite/dashboard/DashboardTeamsOperationsTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/lite/dashboard/DashboardTeamsOperationsTest.kt
@@ -7,9 +7,11 @@ import java.util.concurrent.ConcurrentHashMap
 import okhttp3.OkHttpClient
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.SocketPolicy
 import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNull
 import org.junit.Assert.assertThrows
 import org.junit.Assert.assertTrue
 import org.junit.Before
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 85/100**

Verified against source. The reflected signature `("fetchUserProfile", String, String, StoredCredentials, String)` matches the real private method (DashboardTeamsOperations.kt:489) — Kotlin's `StoredCredentials?`/`String?` erase to the same JVM classes, so `getDeclaredMethod` resolves. And the method genuinely has `catch (_: IOException) { null }` (line 522), so `DISCONNECT_AFTER_REQUEST` makes `execute()` throw → caught → `null`, which means `assertNull(result)` actually exercises the error path rather than short-circuiting on an `InvocationTargetException`. Build + test green.

**Strengths**
- Covers a real, otherwise-hard-to-reach error branch; the description's claim is accurate.
- Correct choice of `SocketPolicy.DISCONNECT_AFTER_REQUEST` to force the `IOException`.

**Concerns**
- Reflection couples the test to the private method's exact name/signature — if `fetchUserProfile` is renamed or reshaped, this fails with `NoSuchMethodException` rather than a clear assertion. Not blocking (the private branch isn't reachable via a public API today), but a comment noting the reflection dependency would help future maintainers.
- **Cross-PR:** #1159 also appends tests to `DashboardTeamsOperationsTest.kt`; both insert before the class's closing brace, so whichever merges second needs a trivial rebase.

**Verdict:** Legitimate, correctly-constructed coverage of the network-failure path — mergeable. (Author == reviewer; second approver needed.)

---

PlataformasInformaticas (approved):

